PR #326, opened 2023-04-13
compute different dt for different parts of V-curve, ionisation volume

FannySkirbekk:
… of the voltage curve in transient mode

## Intent

- [ ] Fix a bug or incorrect behavior.
- [x] Add new capabilities.

## Checklist

- [ ] If relevant, add Sphinx documentation in the rst files. 
- [ ] Add doxygen documentation.

### DIFF
--- a/Physics/StreamerInception/CD_StreamerInceptionStepperImplem.H
+++ b/Physics/StreamerInception/CD_StreamerInceptionStepperImplem.H
@@ -39,14 +39,15 @@ StreamerInceptionStepper<P, F, C>::StreamerInceptionStepper()
   }
 
   // Default settings.
-  m_verbosity          = -1;
-  m_realm              = Realm::Primal;
-  m_phase              = phase::gas;
-  m_mode               = Mode::Stationary;
-  m_transportAlgorithm = TransportAlgorithm::ImExCTU;
-  m_profile            = false;
-  m_debug              = false;
-  m_fullIntegration    = false;
+  m_verbosity           = -1;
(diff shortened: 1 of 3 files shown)


Physics/StreamerInception/CD_StreamerInceptionStepperImplem.H
```diff
@@ -1052,11 +1057,27 @@ StreamerInceptionStepper<P, F, C>::computeDt()
...
+    if(currVoltage == 0){
+      currVoltage = m_voltageCurve(m_time+m_prevDt);
+    }
+    dt = std::abs(m_deltaVPercent*currVoltage/currGradient);
+    
     dt = std::min(dt, m_maxDt);
     dt = std::max(dt, m_minDt);
   }
-
+  m_prevDt = dt;
```
rmrsk, line 1080: Set m_prevDt = 0.0 in constructor for failsafe. 

The advance method does not _have_ to use the dt that is computed here, so m_prevDt should be set to a_dt in the advance method

FannySkirbekk: m_prevDt is set to 0 in the header-file (i.e. initialized to 0 at construction) already. I can change that to 0.0

Ah right, I'll change that now

FannySkirbekk: Or I can change it so it is initialized in the initializer list instead? Up to you

PR #466, opened 2024-06-10
Update example program

rmrsk:
# Summary

Fix a bug in getMaxMinNorm which returned the incorret min-norm. This PR also updates one of the example programs.

### Background

DataOps::getMaxMinNorm had a bug where the min norm was  initialized to numeric_limits<Real>::min() and thus it always returned this value. 

### Solution

Implemented correct initialization.

### Side-effects

None.

### Alternative solutions 

None required.

# Checklist


Exec/Examples/CdrPlasma/DeterministicAir/air_chemistry.json
```diff
@@ -29,8 +29,8 @@
 	     "header": "E/N (Td)\tMobility *N (1/m/V/s)",
 	     "E/N": 0,
 	     "mu*N": 1,
-	     "min E/N": 1,
-	     "max E/N": 10000,
```
rmrsk, line 33: Reset to E/N = [1,10000]

PR #508, opened 2025-04-03
Bug fix for photoionization efficiency in reaction specifier

rmrsk:
# Summary

Fix a bug where ItoKMCPhysics would not use the reaction efficiency when stochastically sampling ionizing reactions.

### Background

ItoKMCJSON would read in a reaction specifier from the JSON file, but when actually reconciling the photoionizing reactions the parent class (ItoKMCPhysics) never used this efficiency. 

### Solution

Added binomial sampling in ItoKMCPhysics where each absorbed photon is subjected to a binomial sampling using the input efficiency and the superphoton weight (which becomes the number of trials).

### Side-effects

There is no safeguard around the binomial sampling. For reaction efficiencies that are either 0 or  1 we could get rid of the sampling entirely.

### Alternative solutions 

Could potentially just use the mean value and/or the variance. 

# Checklist


rmrsk:
Turns out this was supported after all....

PR #540, opened 2025-05-16
Add effective depth for 2D particle depositions

### LINKED ISSUE #539: Potential KMC overflow
Some coarse-grid KMC calculations show numerical overflow when Poisson sampling. Needs to be investigated and proper measures taken -- otherwise we might break with hard charge conservation which is sure to eventually break computer simulations.

rmrsk:
# Summary

This PR supplies two features that were discovered when encountering some overflow issues for 2D Ito-KMC calculations:

1. Permit user-specified depth for particle deposition in 2D.
2. Perform proper roundoff in the ItoKMC method.

TODO: 

- [ ] Add a global variable (perhaps to AmrMesh) that controls the 2D effective depth.
- [ ] CdrPlasma and ItoKMC definitely use volume-scaling, but look into other places where such a scaling is actually taking place (perhaps directly in some solvers?).

Closes #539 

### Background

Particle deposition in 2D assumes grid cells dx=dy and dz=1m. But this can lead to a very large number of particles, and the user should be able to specify dz anyways. In 2D, the user might want to assume symmetry but use dz != 1m. 

Some roundoff issues were discovered in the ItoKMC method -- these triggered because the particles were deposited with effective volumes dx*dy*1m, so it led to a very large number of particles in each grid cell. In addition, several C-style casts were in place instead of static_cast (or llround where appropriate). This PR fixes the primary issues in KMCSolver as well as ItoKMCPhysics to ensure that the real to long long (and opposite) conversions take place with proper casts or rounding.

### Alternative solutions 

# Checklist


rmrsk:
Closing, because I don't think this is a good idea after all....

PR #643, opened 2026-06-05
Compile-time constant BoxLoops for auto-vectorization

### LINKED ISSUE #634: Auto-vectorization hints
BoxLoops should have stronger pragma hints for vectorization. The IntVect abstraction probably hinders auto-vectorization of complex loops.

The solution is probably to use integer-indexing via D_DECL and an explicit stride. Hinting with GCC can be done with -fopenmp-simd (GCC only). But it requires that the user enables this via his/her makefile. Perhaps we should add a vectorization enforcement strategy to the Chombo build system.

rmrsk:
# Summary

Redo BoxLoops::loop into a compile-time constant version. This should expose the guts of the loop fully inlined so that the compiler can more easily vectorize. 

Todo: Use claude setup-token and reset claude.yml and claude-review.yml to use the new OAuth token

Closes #634 

### Background

BoxLoops::loop is an abstraction that takes a box and a functor as an object and runs a dimensionally independent loop over all the cells in that box. Due to the way this is done, it does not always vectorize because of the run-time exposition of the stride. 

### Solution

Redo BoxLoops::loop(Box, Functor, IntVect) to a compile-time version BoxLoop<stride>(Box, Functor). All such loops have been replaced.

Note that vectorization is still not guaranteed -- several follow-up PRs will address this. 

### Side-effects

Quite few, since the loops should be identical. 

### Alternative solutions 

We could write explicit loops, but that's only usable for a few cases where the strided access pattern doesn't fully expose the internals. 

# Checklist

- [x] I have run CheckDocs.py and fixed potentially broken literalincludes
- [ ] I have run a PR review using `@claude review`


rmrsk:
@claude review

rmrsk:
@claude review

rmrsk:
@claude review

rmrsk:
@claude review

rmrsk:
@claude review

rmrsk:
@claude review

rmrsk:
@claude review

PR #655, opened 2026-06-12
Update DataOps to use prebuilt iterators

rmrsk:
# Summary

This PR refactors DataOps (and some other classes) to use prebuilt VoFIterator and FaceIterator when looping over irregular cells. This prevents in-place definitions in favor of prebuilt structures.

### Background

DataOps is a static class that supports various arithmetic operations. The class defined its own iterators when passing through irregular cells, which can cause some overhead and load imbalance. Additionally,some functions would update the cut cells twice, since the regular update would touch the cut cell first and give a result that was correct, only for VoFIterator to come back later and overwrite with the same data.

### Solution 

Change DataOps to take iterators as arguments, which are built and stored in AmrMesh.

### Side-effects

Many changes, with potentially doc and bug inducing characteristics... We have relied on benchmark files to assert correctness.

### Alternative solutions 

None considered.

# Checklist

Before merging the PR:

- [x] I have run clang-tidy and fixed all reported errors.


rmrsk:
@claude review

rmrsk:
@claude review

rmrsk:
@claude review

Issue #609, opened 2026-02-16
Particle merging


rmrsk:
Add an MPI-compliant pair-wise particle merger. Most of the infrastructure will consist of finding pairs and avoiding duplicates. The user should decide how to merge, but with a default merging method available.

rmrsk:
While auditing the `BoxLoops::loop` call sites I profiled-by-inspection the current super-particle merger (`makeSuperparticlesEqualWeightKD`, driven by `ParticleManagement::ParticleMerger<P> = std::function<void(List<P>&, const CellInfo&, int)>`, invoked once per cell). Some findings to fold into the merger redesign:

### Where the cost is today
- **`std::shared_ptr<KDNode>` for the whole tree** (`CD_ParticleManagementImplem.H:226`): every node is a separate heap alloc with atomic refcounting, and `leaves = newLeaves` does a full vector-of-`shared_ptr` copy (atomic bumps) **at every tree level**. Nodes are owned strictly by their parent, so `shared_ptr` buys nothing and costs a lot.
- **Per-cell allocation churn**: the particle vector, `leaves`/`newLeaves`, and each node's internal `ParticleList` are allocated fresh per cell and discarded. Across (cells x merges x steps) this dominates.
- **Particle copying down the tree**: `partitionAndSplitEqualWeightKD` copies sublists into child node vectors at each split.
- The **`std::function` indirection is noise** here — the body is far too heavy for the indirect call to matter. Not worth changing for perf.

### Bug-check finding
In `makeSuperparticlesEqualWeightKD` the timer `t2` ("build_kd") is *declared* but never `CH_START`/`CH_STOP`'d — the build phase (likely the dominant cost) is **currently untimed**. Worth wiring up (and checking the BVH variant) before profiling.

### Interface recommendations (relevant to the redesign)
1. **Stateful strategy object, not a stateless `std::function`.** The win a `std::function` can't give is **holding a reusable scratch arena between calls** (particle vector + a flat KD-node pool). An abstract merger with a `merge(...)` method and member scratch allocates once and `clear()`-reuses.
2. **Hand it contiguous storage, not `List<P>`.** Taking `List<P>&` forces a copy-in/copy-out inside every implementation. Operating on a contiguous span (or the `BinFab` cell storage directly) removes it.
3. **Batch at patch granularity** (`merge(BinFab<P>&, const Box&, ...)`) so scratch is reused across all cells in a patch (thread-local across patches), with `CellInfo` looked up per cell.

[...]

PR #658, opened 2026-06-26
Update cell-wise particle merging

rmrsk:
# Summary

Improves the performance of the cell-wise super-particle merger in `ItoSolver` and adds a new,
opt-in merging algorithm. Addresses the performance findings in #609 (the per-cell allocation
churn in the KD merger); it does **not** implement the MPI pair-wise merger that issue proposes.

### Background

Super-particles are merged per cell (`makeSuperparticles` → the configurable `m_particleMerger`,
invoked once per cell). The equal-weight KD merger built a `std::shared_ptr<KDNode>` tree: every
node was a separate heap allocation with atomic reference counting, each tree level copied a whole
`vector<shared_ptr>` (more atomic traffic), and every node owned its own `std::vector<P>` that
sublists were copied into. Across (cells × merges × steps) this allocation churn and pointer
chasing dominated the merge cost. A standalone benchmark (placed in the git-ignored `Dev/` folder)
confirmed ~160–430 heap allocations *per cell* in the old path.

A latent correctness bug was also found: when the tree reached `maxLeaves` mid-level the build
`break`-ed and discarded the remaining current-level leaves, silently losing their particles
(weight non-conservation) for non-power-of-two `ppc`.

### Solution

**1. Flat-arena KD merger (default, behavior-preserving).** Replace the `shared_ptr` tree with a
reusable, per-thread flat scratch: two ping-ponged contiguous particle buffers plus a flat
`{lo, hi, weight}` range list. Each level streams its children contiguously into the next buffer,
so leaf particles are contiguous spans — no per-node allocation, no atomic refcounting, no pointer
chasing, and the scratch is `clear()`-reused across cells (and across patches via OpenMP threads).
The previously-dead `build_kd` timer is wired up, and the leaf-dropping bug is fixed with a
per-level split budget that carries every remaining node forward (no particle is ever dropped).
Output is **bit-identical** to the previous implementation. The now-unused `CD_KDNode.{H,Implem.H}`
are removed.

Benchmark (both built at `-O3 -march=native`, vs the old implementation): bit-identical output,
weight conserved, steady-state heap allocations **~160–430/cell → 0/cell**, and **~1.7–2.3×**
faster. Profiled by wall-clock, an allocation counter, and `objdump` (the old build's per-node
atomic refcount / `shared_ptr` instructions are gone).

**2. New opt-in merger `sfc_nn` (selectable via `ItoSolver.merge_algorithm = sfc_nn`).** Reaches
the target particle count by space-filling-curve nearest-neighbour clustering, *without* forcing
equal weights:
- `n > ppc`: Hilbert-sort the cell's particles and repeatedly merge the closest adjacent pair down
  to `ppc` (tightest spatial grouping of the methods measured, at the cost of unequal weights);
- `n < ppc`: split the highest-weight particles (co-located halving) up to `ppc`;
- `n == ppc`: identity.
Backed by a reusable, SFC-agnostic `ParticleManagement::mergeAdjacentNearest` primitive (flat binary
heap + index-linked chain, allocation-free; caller supplies the order and a combine callback).
`LoadBalancing::{morton,hilbert}Index` are promoted to `public` so the tested SFC encoders are
reused rather than duplicated.

### Side-effects

- `LoadBalancing::mortonIndex` / `hilbertIndex` change from `protected` to `public` (additive; no
  behaviour change).
- The KD merger's output is bit-identical to before, but the dropped-leaf bug fix means cells that
  previously lost particles (non-power-of-two `ppc` hitting `maxLeaves` mid-level) now conserve
  weight — i.e. results can change in exactly those (previously incorrect) cases.
- `sfc_nn` is a new merging algorithm; selecting it changes results (different grouping, unequal
  weights) and is a quality-oriented option, **not** faster than `equal_weight_kd`. The default
  (`equal_weight_kd`) is unchanged, so existing inputs are unaffected.

### Alternative solutions

- The pure performance angle (replacing the per-node `std::sort` with a weighted quickselect) was
  prototyped and **rejected**: at realistic per-cell counts (e.g. 96 → 16) the sort is not the
  bottleneck, so it gave no measurable speedup. The ~2× win comes entirely from removing the
  allocation churn.
- An equal-weight SFC variant (sort once, cut into `ppc` equal-weight chunks) and an equal-weight
  NN variant were also benchmarked; both are dominated by the flat-arena `equal_weight_kd` on the
  realistic workload, so only the count-based `sfc_nn` (which offers genuinely tighter grouping)
  was kept.

# Checklist

Before merging the PR:


rmrsk:
@claude review

rmrsk:
@claude review

PR #660, opened 2026-06-27
Support variable-sized grid blocks

### LINKED ISSUE #659: Variable-sized tile mesh
Expand TiledMeshRefine to generate varible-sized tiles 8^3, 16^3,... tiles. Use a tree-based merger for this. We might want to do this efficiently since it can lead to millions of tiles, degrading efficiency at scale.

Must rename blocking factor to min_block_size. Realm must have a fast method for returning the level and block containing a point. This is currently in ParticleContainer.

Particle kernels must use a hash grid/tree to find patches. If min/max grid size is the same, we should have a fallback to the current O(1) method.

rmrsk:
# Summary

Implements variable-sized grid blocks for the tiled grid generator (#659): with
`AmrMesh.max_box_size > AmrMesh.min_block_size`, `TiledMeshRefine` emits **variable-sized,
anisotropic boxes** (each a union of aligned `min_block_size` tiles, capped at `max_box_size`), and
**all solvers — including the particle infrastructure — support this**.

Behaviour is **identical to before whenever `min_block_size == max_box_size`** (the default), so this
is a no-op for existing inputs; variable boxes appear only when a user sets `max_box_size >
min_block_size`.

Full design notes are on the issue:
https://github.com/chombo-discharge/chombo-discharge/issues/659#issuecomment-4817109746

Closes #659 

# What the PR does

### Grid generation — top-down over super-tiles
`TiledMeshRefine` generates grids top-down over a grid of *super-tiles* (a `max_box_size` block of
`min_block_size` tiles). Each level is stored compactly as fully-tagged super-tiles (key only) plus,
for partially-tagged ones, a sub-tile bitmask. Fine tiles are materialized only inside partial
(boundary) super-tiles, and only that compact `(key + bitmask)` representation is `Allgatherv`-ed
across ranks — so the intermediate memory scales with the refined **surface + box count**, not the
refined **volume**.

- **Packing:** a full super-tile becomes one box; a partial super-tile is packed by a single shared
  `packTiles()` using a **Berger-Rigoutsos split** (cut at an empty slab in any direction, else the
  strongest signature inflection, else the longest-axis midpoint). Boxes are emitted in sorted-key
  order, so the result is deterministic and **rank-count independent**.
- **Nesting:** the proper-nesting buffer is reproduced exactly from the compact finer-level rep, so
  the grids are **byte-identical to the previous tiler with zero extra cells**.
- `min_block_size == max_box_size` ⇒ every super-tile is a single tile ⇒ one box per tile (legacy).
- A clear abort is raised if `coarsest_domain` is not a multiple of `min_block_size` per direction.

### O(1) point→box lookup for variable boxes
`LevelTiles` registers each box under **every** `min_block_size` tile it covers (`coarsen(box,
minBlockSize)`), keeping the tile→box map uniform for anisotropic boxes; a single-tile box is the
existing fast path. The maps are `std::unordered_map` keyed by a full-component IntVect hash, so the
lookup is genuinely **O(1)** per level (it was `std::map`, O(log n)). Added
`Realm::getLevelAndBox(pos) -> {level, boxIndex, rank, valid}`, the finest-wins lookup the particle
remap uses, exposed for non-particle users.

`ParticleContainer` **aliases the owning Realm's `LevelTiles`** (single source of truth) and never
builds its own. The Realm builds its hash grids during regrid (`regridBase`→`defineLevelTiles`, run
for both the proxy grids in `regridAmr` and the final grids in `regridRealm`), so they are available
downstream — including particle load-balancing, which aliases the proxy hash grid.

### Particles with variable boxes
Removed the `AmrMesh::allocate(ParticleContainer)` guard that forbade particles unless
`max_box_size == min_block_size`. The remaining particle paths are box-shape-agnostic
(`setupGrownGrids` grows arbitrary boxes; the remap "stayer" fast path keys off the per-box valid-cell
mask, not the tile size), and the point→box lookup is handled by `LevelTiles`.

### Rename `blocking_factor` → `min_block_size`
The grid "blocking factor" is really the *minimum* block size once boxes vary, so it is renamed
throughout (option, member, parser, accessor, consumers, all in-repo `.inputs`, docs). The parser
still accepts the old `blocking_factor` key as a **deprecated alias** (`min_block_size` wins), and a
deprecated `getBlockingFactor()` accessor is retained. The Elliptic op factories' operator-internal
`a_blockingFactor` (a Chombo multigrid grid-aggregation parameter, unrelated to the grid tile size) is
intentionally left unchanged.

# Verification

- **Prototype** (gitignored `Dev/`): the top-down generation + exact nesting reproduces the real
  tiler's per-level tile sets **byte-for-byte** for filled / ball / shell patterns, 3-level AMR, and
  refinement ratios **2 and 4**; the packer's coverage is exact (covered cells == tagged cells) and
  Berger-Rigoutsos gives ~20–30% fewer boxes than longest-axis at zero extra cells; peak RSS for the
  top-down intermediate is ~35–65× below the old bottom-up gather at 10M tiles, shape-robust.
- **Real runs** (debug build, assertions on): Electrostatics RodSphere (EB Poisson) 2D + 3D,
  BrownianWalker DriftDiffusion (particles) 2D, Electrostatics MechShaft (multifluid/dielectric) 2D,
  and CdrPlasma RodSphere (coupled EB + Poisson + CDR + radiative transfer) 2D all complete with
  `min_block_size < max_box_size`, and are unchanged at `min_block_size == max_box_size`. The remap
  stayer path's debug self-check passes under variable boxes.
- **Determinism:** np=1 and np=2 produce the identical global grid (verified box/cell counts), so the
  compact gather is order-free and rank-count independent.
- **EB / operators (no code change):** `min_block_size` is used only for `LevelTiles`/the lookup; EBIS
  uses the independent `max_ebis_box`; the elliptic/AMR operators are box-based.
- Sphinx (`-W`) and `doxygen-check` pass; all pre-commit hooks pass on every commit.

# Design decisions

1. **Anisotropic, deterministic boxes via KD / Berger-Rigoutsos bisection over the tile grid** — not an
   octree, which assumes a cubic / power-of-two index space, whereas domains are only required to be a
   multiple of `min_block_size` and are routinely non-cubic / non-power-of-two.
2. **O(1) min-tile hash-grid lookup** — each box registered under all its min-tiles; `min == max` is the
   one-entry-per-box fast path. (A per-level box BVH is a future memory fallback; not built.)
3. **Top-down generation with exact dilation nesting** — chosen over a conservative whole-super nesting
   because it keeps grids byte-identical and adds zero cells.

Trade-off accepted: per-super packing does not merge boxes across super-tile boundaries, so it produces
~11–21% more boxes than a hypothetical global merge of the whole level — but the global merge needs the
full fine-tile set (the memory being eliminated), and the BR split more than recovers that gap, at zero
cell cost.

# Open follow-ups (not in this PR)

- EB load metric: large/elongated boxes can concentrate cut cells; weighting cut cells in load balancing
  is a measurement-driven follow-up.
- Possible rename of `grid_algorithm` values `br`/`tiled` (orthogonal; would touch all inputs again).

# Checklist

Before merging the PR:


rmrsk:
@claude review

rmrsk:
@claude review

PR #661, opened 2026-06-28
Geometric multigrid improvements

rmrsk:
### Background

This branch collects two related lines of work on the geometric multigrid (GMG) Helmholtz solver:

1. **Correctness fixes** to the GMG operator (single-phase `EBHelmholtzOp` and multi-phase `MFHelmholtzOp`) and the EB gradient.
2. A new, first-class **multi-solver fallback chain** for the elliptic solvers — stand-alone GMG plus custom GMRES/BiCGStab Krylov solvers preconditioned by the GMG cycle — that converges robustly on production problems where stand-alone GMG stalls, and is written to scale to large core counts.

### GMG / EB operator bug fixes

- **`dotProduct` κ² weighting:** The irregular-cell inner product accumulated `(κ·X)·(κ·Y)` instead of `κ·X·Y`. Fixed in both `EBHelmholtzOp` and `MFHelmholtzOp`.

- **Domain BC diagonal weight:** `computeDiagWeight` applied a unit multiplier to all domain faces regardless of BC type. `EBHelmholtzDomainBC::getDiagWeight()` is now a **pure virtual** so every domain BC declares its diagonal contribution explicitly: Dirichlet `2`, Neumann `0`, Robin `1` (Larsen inherits Robin), and the composite Electrostatic and Eddington BCs route to their per-face sub-BC. The base method's documentation explains how the weight is chosen (the coefficient of `φ_i` in `getFaceFlux`, which it must stay consistent with). The weight feeds only the relaxation/smoother diagonal, so an inconsistent value affects convergence rate, never the converged solution.

- **`fillDomainFlux` homogeneous BC:** The method hardcoded `false` for the physical BC flag. The flag is now threaded through from `refluxFreeAMROperator`.

- **Double exchange in `MFHelmholtzOp`:** Six delegation methods performed an MF-level ghost exchange and then let the per-phase `EBHelmholtzOp` repeat it. Added `turnOffExchange()`/`turnOnExchange()` guards around all per-phase calls.

- **`matchBC` VoF aliasing:** `MFHelmholtzJumpBC::matchBC` read from `VolIndex(iv, 0)` as a proxy for the first VoF in the opposite phase. Replaced with `vofsPhase1[0]` / `vofsPhase0[0]`.

- **EBGradient at EBCF cells:** the reconstructed gradient was garbage at embedded-boundary / coarse-fine cells when `refRat > num_ghost`. Fixed, and a stale precision comment was removed.

### Reflux-free AMR operator flag

- Added `gmg_reflux_free` (default `false`) to `FieldSolverGMG`, `EddingtonSP1`, `CdrGodunov`, and `CdrCTU`. When enabled, `EBHelmholtzOp::AMROperator` routes to `refluxFreeAMROperator`, which fills coarse-level CF-interface fluxes by conservatively averaging fine-level fluxes computed from the composite solution. The two formulations are mathematically equivalent.

### Additional multigrid smoothers and cycles

- Added a **Chebyshev** polynomial smoother and a **restricted additive Schwarz (RAS)** block smoother, selectable via `gmg_smoother` (`chebyshev <order> <eig_ratio>`, `ras`); the RAS inner-sweep count is threaded through the solver/factory/operator chain.
- Exposed **W-cycles** via `gmg_cycle = wcycle` (in addition to `vcycle`).

### Multi-solver fallback chain (new solver path)

Stand-alone GMG occasionally stalls on hard production problems (strong permittivity contrast, awkward geometry). Wrapping the cycle in an outer Krylov method converges robustly there. Each multigrid-based solver (`FieldSolverGMG`, `CdrCTU`/`CdrGodunov`, `EddingtonSP1`) gains a top-level solver selector that accepts a fallback chain:

```
<Solver>.solver                = gmg gmres            # space-separated chain, tried in order
<Solver>.krylov_eps            = 1.E-10               # outer Krylov relative tolerance
<Solver>.krylov_max_iter       = 50
<Solver>.krylov_restart        = 30                   # GMRES only
<Solver>.krylov_vcycles        = 1                    # multigrid cycles per preconditioner application
<Solver>.krylov_retry_interval = 1                    # re-probe GMG every N solves while latched to the fallback
<Solver>.krylov_reprobe_iters  = 0                    # adaptive early re-probe (0 = off)
```

- `gmg` (the previous behaviour) is stand-alone cycling; `gmres`/`bicgstab` use the multigrid cycle as a **preconditioner**. The `gmg_*` keys configure that cycle in all cases.
- A space-separated list of any length is a **fallback chain** tried in order (e.g. `gmg gmres`, `gmg bicgstab gmres`): cheap multigrid runs first, and the chain falls back to a Krylov method only on the steps where multigrid stalls.
- **Warm-start rule:** each attempt warm-starts from the previous attempt's result *only if* that result reduced the residual below the chain's initial residual; otherwise it restarts from the initial-residual solution, so a diverged attempt never pollutes the next solver's initial guess.
- **Re-probe latch** (`krylov_retry_interval`): when a chain beginning with `gmg` fails *persistently*, the chain latches onto the Krylov fallback and only re-probes GMG every N solves (the latch state survives regrids), instead of wasting a full failed cycle batch on every solve before the fallback even starts.
- **Adaptive re-probe** (`krylov_reprobe_iters`): while latched, re-probe GMG early whenever the Krylov solver converges in ≤ this many iterations — a strong preconditioner suggests GMG alone may now succeed.

#### Custom, scale-out Krylov solvers

The outer Krylov solvers are **written in chombo-discharge** (rather than wrapping Chombo's `GMRESSolver`/`BiCGStabSolver`) so the path scales to large core counts. New header-only infrastructure in `Source/Elliptic/`:

- **`AMRMultigridKrylovOp<T>`** — a `LinearOp` over `Vector<LevelData<T>*>` that reuses an already set-up `AMRMultiGrid` for the matvec (`computeAMROperator`) and the preconditioner (`AMRVCycle`). The AMR inner product is allocation-free and reuses the Realm's existing valid-cell mask (`AmrMesh::getValidCells`); per-level partials are reduced in a **single batched `MPI_Allreduce`** (`dotProduct` and a new `mDotProduct`), and the geometry-only cell volume is cached across solves. `norm(.,2)` is the inner-product-induced norm GMRES requires.
- **`KrylovGMRES<T>`** — restarted, right-preconditioned GMRES(m) with a persistent Arnoldi basis (allocated once per regrid, not per solve) and classical Gram-Schmidt + one reorthogonalization, issued through the batched `mDotProduct` so each Arnoldi step costs a single reduction regardless of the AMR level count.
- **`KrylovBiCGStab<T>`** — BiCGStab with eight persistent work vectors and a fused `omega` reduction.
- **`EllipticSolverChain::KrylovSolver<T>`** — the residual-correction driver and owner of the adapter, the two Krylov solvers, and the fallback-latch policy. Warm-start aware: it folds the inhomogeneous boundary / dielectric surface-charge jump into the initial residual so the matvec and preconditioner run with homogeneous BCs.

Together these replace the per-solve allocation churn and per-level reduction storm of the original prototype: the work vectors are reused across solves, and per-iteration reductions drop from `O(L)` / `O(iter·L)` allreduces to a small constant independent of the AMR level count `L`.

Correctness issues fixed while promoting the prototype to this path (each verified against the stand-alone `gmg` result):

- The driver must **not** negate `computeAMRResidual` — Chombo already returns the residual as `rhs - L(phi)`. Negating it flipped the correction sign, producing a wrong field on warm starts (`phi = 2·phi0 - phi*`); the cold case survived only because `-phi*` has the right field magnitude, which is why convergence-only checks missed it.
- The cycle preconditioner calls `AMRVCycle` directly, which needs the multigrid solver's *intermediate* depths. The host solvers `revert()` after every solve, so the driver re-`init()`s before each Krylov solve (mirroring `AMRMultiGrid::solve`); otherwise the 2nd+ solve on a fixed grid runs the cycle with reverted (full-coarsening) depths.
- Removed a redundant matvec per Krylov iteration in the preconditioner (the residual recompute is only needed for `krylov_vcycles > 1`).

#### Interface and output

- The `solver` and all `krylov_*` keys are **mandatory** — parsed with `pp.get` in each host's `parseMultigridSettings` (like the `gmg_*` keys) and present in every `.inputs`/`.options` that drives one of these solvers.
- The shipped default in all in-repo option/input files is now **`solver = gmg gmres`**: cheap multigrid with an automatic GMRES fallback. (`solver = gmg` still reproduces the previous stand-alone behaviour exactly.)
- The Krylov solver **reuses `gmg_verbosity`** (no separate `krylov_verbosity`). At `gmg_verbosity >= 1`, a one-line convergence summary is printed per solve, and the host prints a message whenever the chain switches solver. The printed residuals are **relative residuals** — the true residual normalized by the zero-residual `||rhs - L(0)||` — so progress is on a common, solver-independent scale directly comparable to `krylov_eps`. The GMRES/BiCGStab per-iteration history appears at `gmg_verbosity >= 3`.

**Documentation:** `LinearSolvers.rst` is the canonical reference — a "Krylov-accelerated multigrid" narrative plus a "Tuning multigrid performance" chapter (label `Chap:MultigridTuning`) that documents every `gmg_*` / `solver` / `krylov_*` key, the new smoothers, and the fallback options, and clarifies that `gmg_bottom_solver` (the coarse-grid solver inside each cycle) is distinct from the new `solver` key. `Electrostatics.rst` and `RTE.rst` cross-reference it (a previously duplicated tuning chapter and label were consolidated here).

### Side-effects

- **`EBHelmholtzDomainBC::getDiagWeight` is now pure virtual:** any out-of-tree domain-BC subclass must implement it (it previously had a `1.0` default).
- `fillDomainFlux` gains a `bool` parameter; any out-of-tree caller must be updated.
- **Not backward compatible:** the `solver` and `krylov_*` keys are mandatory. Input scripts that drive `FieldSolverGMG`, `CdrCTU`, `CdrGodunov`, or `EddingtonSP1` must now list them; all in-repo examples and tests are updated. The previously experimental `gmg_krylov_accel` boolean is removed.
- The default `solver = gmg` reproduces the previous solver behaviour exactly.

### Alternative solutions

N/A.

## Checklist

Before merging the PR:


rmrsk:
@claude review

PR #662, opened 2026-06-30
Declarative H5Part column descriptor for SoA particle plotting (#657)

### LINKED ISSUE #657: Declarative named-subset H5Part column descriptor for SoA particle plotting
## Summary

Replace the hand-maintained, positionally-coupled H5Part variable-naming scheme used for SoA
particle plotting with a declarative, **named-subset** column descriptor in `ParticleTraits`,
analogous to the existing `h5Columns` checkpoint mechanism.

## Current situation

H5Part plotting of SoA particles (`DischargeIO::writeH5Part`) requires the caller to pass the
payload datasets as explicit `(name, accessor)` pairs. Today the only consumer,
`ItoKMCGodunovStepper::plotParticles`
(`Physics/ItoKMC/TimeSteppers/ItoKMCGodunovStepper/CD_ItoKMCGodunovStepperImplem.H`), builds
those pairs by **positionally indexing** two free-standing string vectors against a parallel,
hand-written list of accessor lambdas:

```cpp
std::vector<std::string> ItoParticle::s_realVariables = {"weight", "mobility", "diffusion", "energy", "scratch"};
std::vector<std::string> ItoParticle::s_vectVariables = {"old", "v", "scratch"};
```

```cpp
{ItoParticle::s_realVariables[4], [](const Leaf& l, std::size_t i) { return l.get<&ItoParticle::scratch>(i); }},
...
```

This has several problems:

- **Three sources of truth that must agree by hand**: the `s_realVariables`/`s_vectVariables`
  strings (defined in `CD_ItoParticle.cpp`), the accessor lambda list in the stepper, and the
  `ParticleTraits<ItoParticle>::columns` member-pointer tuple. Reordering or renaming any one
  silently mislabels datasets, or indexes out of range.
- `CD_ItoParticle.cpp` exists **solely** to provide the out-of-line definition of those two
  `static std::vector<std::string>` members. They are read in exactly one place.
- The naming logic is duplicated per payload type rather than living with the type, even though
  there are five `ParticleTraits` specializations (`ItoParticle`, `Photon`, `TracerParticle`,
  `ItoKMCFieldParticle`, `DischargeInceptionParticle`).

## Proposed design

Introduce an optional, declarative descriptor in `ParticleTraits` listing the **subset** of
payload columns that should appear in H5Part output, each carrying its dataset name and its
scalar/vector grouping — mirroring how `h5Columns` already declares the checkpoint subset:

```cpp
template <>
struct ParticleTraits<ItoParticle>
{
  static constexpr auto columns = std::make_tuple(/* all payload columns, unchanged */);

  // Checkpoint subset (already exists).
  static constexpr auto h5Columns = std::make_tuple(D_DECL(&ItoParticle::vx, &ItoParticle::vy, &ItoParticle::vz));

  // NEW: H5Part plotting subset — only SOME members, each named + grouped.
  static constexpr auto h5PartColumns = std::make_tuple(
    Scalar{"mobility",  &ItoParticle::mobility},
    Vector{"v", D_DECL(&ItoParticle::vx, &ItoParticle::vy, &ItoParticle::vz)});
};
```

A single generic `writeH5Part(filename, container, shift, time)` overload then walks
`h5PartColumns`, emitting each `Scalar` as one dataset and each `Vector` as `name-x/name-y/name-z`,
reading values generically via the member pointers (the SoA already provides `get<MemberPtr>`).
Container-owned `position` and `weight` are emitted unconditionally, as today.

### Key requirement: named **subset**, not all members

H5Part output must **not** auto-include every payload column. `scratch`, `old`, etc. are internal
scratch/bookkeeping fields and should not be plotted. The descriptor is an opt-in allow-list with
explicit names — exactly the `h5Columns` philosophy — so each payload type chooses which fields are
visualized and under what name.

## Benefits

- Single source of truth: the name lives next to the member pointer; reorder freely (matches the
  name-based-access design note in `CD_ParticleSoA.H`).
- Delete `s_realVariables`, `s_vectVariables`, and therefore `CD_ItoParticle.cpp` entirely.
- Collapse the ~40-line `scalarVars`/`vectorVars` lambda table in the Godunov stepper to a single
  call.
- Generic across all payload types; any future plotting (e.g. `McPhoto`/`Photon`) gets it for free.

## Constraints / notes

- No member-name reflection before C++26, so the dataset string is still written by hand once —
  but co-located with the member pointer so it cannot drift positionally.
- The existing flat `columns` tuple flattens `D_DECL` vector triples inline, so it cannot by itself
  recover scalar-vs-vector grouping; `h5PartColumns` must encode the grouping (via the `Scalar`/`Vector`
  descriptors above).
- Touches the `ParticleTraits` contract; the five existing specializations migrate mechanically.
- Changes HDF5 dataset names to whatever `h5PartColumns` declares (same output-naming consideration
  as renames of the payload members).

## Scope

Separate PR from the in-flight `ItoParticle` SoA member rename (`particle_soa` branch); this issue
records the design for that follow-up.

rmrsk:
# Summary

Implements the declarative H5Part column descriptor proposed in #657.

Closes #657 

### Background

H5Part particle output previously relied on hand-maintained, position-dependent lists: the caller passed two parallel `std::vector<std::pair<std::string, lambda>>` arguments to `DischargeIO::writeH5Part`, and for `ItoParticle` the dataset names lived in separate `s_realVariables`/`s_vectVariables` string vectors whose ordering had to be kept in lock-step with the accessor lambdas. This was duplicative, error-prone (positional coupling between a name and the field it names), and exposed internal scratch columns unless the caller manually omitted them. Only one call site (the ItoKMC Godunov stepper) actually used it.

### Solution

Introduce an optional, declarative `h5PartColumns` descriptor in `ParticleTraits`, so the dataset names co-locate with the payload member pointers — a single source of truth, opt-in, with no positional coupling.

- Add `H5Part::Scalar` / `H5Part::Vector` descriptor types in `CD_ParticleSoA.H` (placed in a nested `H5Part` namespace so `H5Part::Vector` does not collide with Chombo's `Vector`). Each descriptor pairs a dataset name with the pointer(s)-to-member selecting the column(s):

  ```cpp
  static constexpr auto h5PartColumns = std::make_tuple(
    H5Part::Scalar{"mobility", &ItoParticle::mobility},
    H5Part::Vector{"v", D_DECL(&ItoParticle::vx, &ItoParticle::vy, &ItoParticle::vz)});
  ```

- Rewrite `DischargeIO::writeH5Part` to take only `(filename, particles, shift, time)`. `id`, position and `weight` are always written automatically; payload datasets are derived by compile-time introspection of `Traits::h5PartColumns` (each `H5Part::Vector` is written as `name-x`/`name-y`/`name-z`). Particle types that declare no `h5PartColumns` (or an empty one) get `id`/position/`weight` only.
- Declare `h5PartColumns` for the particle types: `Photon` (`kappa`, `v`), `ItoParticle` (`mobility`, `diffusion`, `energy`, `v`), `TracerParticle` (`v`), `ItoKMCFieldParticle` (none — transient internal type), `DischargeInceptionParticle` (`v`).
- Remove `ItoParticle::s_realVariables`/`s_vectVariables` and the now-empty `CD_ItoParticle.cpp`; simplify the ItoKMC Godunov stepper call site to the single-argument-list call.
- Update the Sphinx particle-visualization documentation (new signature + descriptor example).

### Side-effects

- **API change:** the lambda-list overload of `DischargeIO::writeH5Part` (and the `H5PartScalar`/`H5PartVect` aliases) is removed in favour of the descriptor-driven overload. The only in-tree caller (ItoKMC Godunov stepper) is updated.
- **Behaviour change for ItoKMC particle plots:** internal scratch/previous-position columns (`scratch`, `old`, `scratch_*`) are no longer written; the velocity vector dataset is now named `v` (`v-x`/`v-y`/`v-z`). `weight` is still written automatically.
- `CD_ItoParticle.cpp` is deleted (ItoParticle becomes header-only, like the other payload types).
- The new `writeH5Part` also explicitly closes the two HDF5 dataspaces it creates (the previous version leaked them).

### Alternative solutions

- Keep the lambda-list overload alongside the descriptor for arbitrary computed quantities. Dropped because no in-tree caller needs computed quantities and #657 explicitly aims to eliminate the duplication; it can be re-added later if a real need appears.

# Checklist

Before merging the PR:


rmrsk:
@claude review

rmrsk:
@claude review

PR #663, opened 2026-06-30
Support for filling ghost particles

### LINKED ISSUE #642: Fill ghost particles
Plans are in place for a new particle-merging strategy, but this strategy requires us to fill ghost particles.

We must extend ParticleContainer with both the appropriate members and functionality for fetching ghost particles. This is a bit more complex than it sounds....

Verification tests should include load-up of a pre-defined particle distribution that is put on the AMR mesh. After filling the ghost particles, we should go through the pre-defined distribution to ensure that we have, indeed, filled all the relevant ghost particles. For the record, this needs to be scalable, so we might need to move particles onto grown coarsened/refined grids for locality, and then fill.

rmrsk:
## Summary

This PR adds particle **ghost-filling** to `AmrMesh`/`Realm` and `ParticleContainer`. A *ghost particle* is a transient, non-owned copy of a particle that lives in an adjacent region and overlaps a patch's ghosted/grown box. Copies can come from three directions relative to the receiving patch — a same-level neighbour, the next-coarser level, or the next-finer level — and live **in the same `ParticleSoA` leaf** as the valid particles, marked in place so per-patch code can tell them apart via `ParticleSoA::isGhost()`.

The user chooses the ghost **width** (a minimum, measured in destination cells) and can independently enable same-level, coarse-to-fine, and fine-to-coarse filling, as shown below.

image

Closes #642

## ParticleSoA

- Mandatory 1-byte `enum class GhostType : std::uint8_t { Valid, SameLevel, Coarse, Fine }` column, alongside `particleID`/`rankID`, with `ghost(i)`/`isGhost(i)` accessors. The enumerators record provenance.
- The per-column index-sequence machinery threads it through allocation, move, bulk `append`/`catenate`, swap-pop `remove`, and MPI linearization automatically; the HDF5 subset excludes it (never checkpointed). `s_invalidID` was made public.

## Ghost masks (`Realm` / `AmrMesh`)

- Per-direction **prebuilt masks** (`ParticleGhostMask`: per-box CSR of source cell → destination boxes), rebuilt at every regrid, one collection per direction (`AMRParticleGhostMask`).
- Registered by width so only needed widths are built: `AmrMesh::registerParticleGhostMask(realm, width)`; fetched with `getParticleGhostMask` / `getParticleGhostMaskCoarToFine` / `getParticleGhostMaskFineToCoar`; `getTrivialParticleGhostMask()` opts a direction out.
- Built from Copier motion-plans (only overlapping box pairs), restricted to the coarse-fine halo, and OpenMP-parallel over grid boxes.
- **Coarse-to-fine over-scatter fix**: coarse particles are stored per coarse cell but the coarse-to-fine ghost width is measured in fine cells, so a whole boundary coarse cell would over-communicate by up to the refinement ratio. Each coarse-to-fine target now carries an *acceptance box* (the fine box's ghost region, periodic-shifted), and the scatter keeps a coarse particle only if its fine cell lands in it — pruning to the exact fine shell. Same-level and fine-to-coarse are already exact.

## ParticleContainer

- **`fillGhost(same, c2f, f2c)`** — for every valid particle, scatters a copy (same-rank directly, cross-rank via MPI) into each destination leaf listed by the masks, tagged with the receiver-view `GhostType`, keeping the source's `particleID`. Clears the halo first; pass the trivial mask to skip a direction.
- **`clearGhostParticles()`** — swap-pop removal of every ghost. Ghosts are transient: dropped by `remap()`/`preRegrid()` and excluded from `getNumberOfValidParticlesLocal()`.
- **`resetRankIDs(onlyValidRegion)`** and **`resetParticleIDs(onlyValidRegion, startID) -> count`** — bulk (re)assign owner rank / globally-unique contiguous ids to owned particles (ghosts always skipped so their copied identity survives); the valid-region test is purely geometric on the patch box; uniqueness via an all-gathered per-rank exclusive prefix.

## Utilities

- `ParallelOps::gather(long long)` overload (used for the id prefix sum).

## Verification

- **2D and 3D, OPT, serial and `mpirun -np 4` (bit-identical between them)**: `fillGhost`, the three masks, and the coarse-to-fine prune.
- The coarse-to-fine ghost count drops ~4.5× (refinement ratio 4) with the acceptance-box prune, while same-level and fine-to-coarse counts are unchanged.
- `resetParticleIDs` produces a permutation of `[0, n)` (count and sum checks) with zero wrong ranks, in 2D and 3D.
- The verification stepper dumps the per-direction ghost particles to H5Part files for visual inspection.
- Docs: new **Ghost particles** section in `Docs/Sphinx/source/Source/Particles.rst` (masks + `fillGhost` signature + figure); `doxygen-check` and Sphinx `-W` build clean.

### PR-review checklist

Mandatory:

Optional:


rmrsk:
@claude review

PR #669, opened 2026-07-29
Add a distributed nearest-neighbor super-particle merge

### LINKED ISSUE #609: Particle merging

rmrsk:
# Summary

Adds a distributed, MPI-safe nearest-neighbour **pair** super-particle merge (`merge_algorithm = nn_pair`) and, around it, reorganizes the `ItoSolver` particle-merge interface and the ItoKMC merge timing so that per-cell and AMR-wide merge methods are handled uniformly.

Closes #609 

### Background

The existing super-particle mergers (`equal_weight_kd`, `reinitialize`, `nn_sfc`, …) all operate strictly *within a single grid cell*. That biases particles near patch/rank boundaries toward worse matches than interior ones, and can never merge a genuinely nearest pair that straddles a boundary. `nn_pair` finds each over-crowded particle's *true* nearest neighbour across patch and rank boundaries.

Wiring an AMR-wide (whole-container, cross-rank) merge into `ItoSolver`/ItoKMC also exposed that the merge machinery was hard-wired to per-cell operation, which the rest of this PR relaxes.

### Solution

**The merge itself** — `ParticleManagement::mergeNearestNeighborsRound` (`Source/Particle/CD_NearestNeighborParticleMerge.H` / `…Implem.H`, `RULES.md` is the design spec):
- Candidate search uses EBGeometry's SIMD `PointCloudBVH` with a batched `allNearestNeighbors` k-NN graph; a propose/judge/verdict protocol resolves cross-patch/cross-rank pairs safely (any number of simultaneous suitors), with deterministic tie-breaks. One round does a single cross-patch exchange.

**`ItoSolver::makeSuperparticlesNnPair`** turns that into a complete *reach-the-target* algorithm:
- **Drain loop**: one round merges nearest-neighbour *pairs*, so a round roughly halves an over-full cell's surplus; the round is repeated until the global count stops decreasing.
- **Splitting**: a cell *below* target is brought up to it by a per-cell greedy heaviest-split (same rule as `nn_sfc`).
- **Minimal particle**: the merge runs on a small energy-only `ItoMergeParticle` container (extract → kill ItoParticles → merge+split → rebuild), so the repeated cross-rank ghost exchange and container churn move ~2 reals/particle instead of the full `ItoParticle`.
- **Configurable**: `nn_pair_iterate` / `nn_pair_fallback` / `nn_pair_max_cell_dist` inputs.

**Interface / dispatch:**
- `ParticleManagement::ParticleMergeKind{Cell,AMR}` and `ParticleMergeMethod{None,EqualWeightKD,Reinitialize,ReinitializeBVH,NnSfc,NnPair,External}`. `ItoSolver` stores the method and derives the granularity.
- `makeSuperparticles(container, int|Vector<int>)` is now a **router** over the method; every method is also a **public, directly-callable** `makeSuperparticlesXxx()`. Each cell method builds its per-cell merger *locally* and applies it through a shared `applyCellMerger()`; `m_particleCellMerger` survives only for `merge_algorithm = external`.
- **ItoKMC merges after the chemistry advance** through the public `makeSuperparticles()` (previously folded into `reconcileParticles`' per-cell kernels only because the particles were cell-sorted there). This fixes a latent gap: `nn_pair` previously reduced to the no-op per-cell hook and did not merge during a step.

**Supporting changes:**
- `Realm::defineParticleGhostMasks()` moves from `regridBase()` to the operator/mask phase (`regridOperators()`) so a width registered in a solver's `registerOperators()` is honoured on the first grid build.
- `ParticleContainer::findDestination()` (and its `Destination` struct) return to `public` (the merge locates merged particles via the same tile-hash query `remap()` uses, not a linear patch scan).
- `ParticleManagement` gains an authoritative merge-algorithm catalog + a `CD_ParticleMergers.H` umbrella; the merge headers are split by dependency weight so the ~40 light consumers of the per-cell mergers don't transitively pull in EBGeometry/AmrMesh/MPI.
- Selector rename `sfc_nn` → `nn_sfc` (nn_ nearest-neighbour family).
- `Exec/Tests/ItoKMC/NNMergeSmokeTest`: RodNeedleDisk discharge exercising the production path.

### Side-effects

- `merge_algorithm` selector `sfc_nn` → `nn_sfc`; new selector `nn_pair`. Any input file using `sfc_nn` (or `nearest_neighbor`, which never reached main) must update.
- `ItoSolver`: the internal `makeSuperparticles(.,.,level[,dit])` overloads and `mergeParticles()` are removed; new public `makeSuperparticlesXxx()` per method.
- ItoKMC per-step merge now runs after the chemistry advance rather than inside the per-cell reconciliation (equivalent result for cell methods; `nn_pair` now actually merges per step). Costs one extra `organizeParticlesByCell` per merge step.
- `Realm::defineParticleGhostMasks()` runs one regrid-phase later (no behavioural change — nothing else registers a particle ghost mask width).
- `ParticleContainer::findDestination()`/`Destination` change from `protected` to `public`.

### Alternative solutions

Keeping the search per-cell (rejected — the boundary-bias problem is exactly what this addresses). A hand-rolled spatial index was prototyped and dropped in favour of EBGeometry's `PointCloudBVH`. Merging the full `ItoParticle` directly is retained in git history; whether the minimal-particle path is a net win wants profiling on a crowded, many-round case.

### PR-review checklist

Mandatory:

- [x] I have run CheckDocs.py and fixed potentially broken literalincludes. *(Also verified with a local `sphinx-build -W`.)*
- [x] I have added all relevant user documentation to Sphinx. *(Ito.rst: `nn_pair`/`nn_sfc`, options, merge/split behaviour.)*

Optional:


rmrsk:
@claude review

rmrsk:
@claude review

rmrsk:
@claude review

rmrsk:
@claude review

PR #673, opened 2026-07-31
Disable namespace-body indentation and adopt leading-star Doxygen comments

### LINKED ISSUE #671: Namespace indentation
## Proposal

Disable namespace-body indentation in `chombo-discharge`. The `.clang-format` at the repo root currently sets:

```yaml
NamespaceIndentation: All
```

We should change it to:

```yaml
NamespaceIndentation: None
```

and reformat the tree once so no code is indented merely because it sits inside a namespace.

## Why

Every namespace level currently adds two spaces of indentation to its entire body. chombo-discharge nests namespaces (the Chombo/`ChomboDischarge` namespace opened by `CD_NamespaceHeader.H`, then module namespaces like `ParticleManagement`, and now a `detail` sub-namespace), so real code starts several columns in before any actual scope, eating horizontal budget on an 80/120-col style.

More importantly, `NamespaceIndentation: All` makes namespace structure expensive to change:

- **Introducing or removing a namespace re-indents its whole body.** Wrapping a block of declarations in a new nested namespace (e.g. moving internal helpers into `ParticleManagement::detail`) forces clang-format to shift every line inside it by two spaces, producing a large mechanical whitespace diff that swamps the actual change and makes review harder.
- **It interacts badly with our Doxygen comment style.** Our block comments use no leading `*`:

  ```cpp
  /**
    @brief ...
    @details ...
  */
  ```

  When the enclosing indent changes, clang-format re-indents the `/**` line and the code but leaves these comment interiors where they were, so the interiors and the closing `*/` end up misaligned relative to `/**`. clang-format neither fixes this nor flags it (the result is "clean" to `--dry-run`), so the misalignment has to be repaired by hand after every namespace move.

Setting `NamespaceIndentation: None` fixes both: code sits at a stable indentation regardless of how deeply it is namespaced, adding/removing a namespace is a near-zero diff (just the `namespace X {` / `}` lines), and the comment-mangling disappears because namespacing no longer re-indents anything.

## Scope / cost

One-time, purely mechanical reformat of the tree (`pre-commit run clang-format --all-files` after the `.clang-format` change), landed as its own commit so it is easy to review and to skip in `git blame` (add it to `.git-blame-ignore-revs`). No source or behavioral changes.

## Context

This came up while moving the nearest-neighbor merge's internal helpers/types into `ParticleManagement::detail`: the namespace wrap alone generated a ~900-line whitespace diff and required hand-realigning every wrapped comment interior. `NamespaceIndentation: None` would have made that change a handful of lines.

rmrsk:
## Summary

- Set `NamespaceIndentation: None` in `.clang-format` (was `All`) and reformat the tree once with the pinned clang-format. chombo-discharge nests namespaces (the `CD_NamespaceHeader.H` namespace, module namespaces, and now `detail` sub-namespaces), so `All` was eating horizontal budget on an 80/120-col style and turning every namespace wrap/unwrap into a large, hard-to-review whitespace diff. With `None`, code sits at a stable indentation regardless of namespace nesting depth, so adding/removing a namespace is a near-zero diff.
- Convert all multi-line `/** ... */` Doxygen blocks to a leading-`*`-per-line style (one space between `*` and the following text) and enable `ReflowComments: true`. Our previous no-leading-star convention meant clang-format treated comment-block interiors as opaque text — it could never detect or repair indentation drift (e.g. from namespaces that used to nest more deeply), and `ReflowComments` had nothing to work with. With a leading `*` on every line, clang-format now owns these blocks: it reflows overlong `@details`/`@param` prose (and plain `//` comments) to the column limit and keeps continuation lines correctly indented under the `*`, so this can't drift again.
- Fix every `Docs/Sphinx` `literalinclude` directive whose `:lines:`/`:dedent:`/`:emphasize-lines:` had drifted out of sync with the reformatted source (line counts and indentation both shifted). Some drift caused loud Sphinx warnings (bad dedent, unparsable excerpts); some was silent — a still-valid-looking but wrong excerpt with no warning at all. Every directive was re-resolved mechanically: unique marker comments were inserted into a pristine pre-reformat copy of the tree at each directive's exact boundaries, that copy was put through the identical reformatting pipeline, and the markers' new positions were read back to compute corrected values — avoiding the ambiguity of fuzzy text/word matching once code is rewrapped. `sphinx-build -W --keep-going` now completes with 0 warnings (was 46 after the first two commits), matching `main`.
- Update `CLAUDE.md`'s Doxygen style section (and file-level example) to document the new leading-star convention, and add a section explaining why `literalinclude` directives need re-checking after any mass C++ reformat, plus the marker-based technique for fixing them.
- **(Review feedback)** Protect ~100 hand-formatted `//` comment blocks (hanging-indent "TLDR:"-style prose, ASCII-art diagrams, numbered/lettered lists, formula listings across ~35 files) from `ReflowComments`, which mishandles all of them the same way — it groups comment lines into paragraphs by the whitespace after `//`, so any block that intentionally varies that spacing gets split into independently-reflowed fragments. Each block was restored to its exact original text and wrapped in `// clang-format off`/`// clang-format on` rather than auto-flattened, since several are lists/diagrams where flattening into prose would be actively wrong. Also converges clang-format to a stable fixed point on the whole tree (repeated runs now produce zero further changes).

The first two changes have to land together: the namespace-indentation reformat alone shifts code left without touching hand-written comment bodies, which briefly makes existing indentation drift more visible; only combining it with the comment-reflow change actually leaves the tree in a consistent, correctly-indented state end to end. The third commit is a required correctness fix for the same reformat, since without it the `Build-documentation` CI job (which runs Sphinx with warnings-as-errors) would fail, and worse, some doc excerpts would silently show the wrong code with no warning at all.

Fixes #671

## Test plan

- [x] `pre-commit run --all-files` passes (clang-format, reuse, codespell, cppcheck, format-input-files, doxygen-check)
- [x] `git diff --ignore-all-space` confirms commit 1 is whitespace/pure-rewrap only
- [x] Comment-content-preservation check (decoration/whitespace-normalized diff) confirms commit 2 changes no wording
- [x] Manually spot-checked `@code` blocks (`CD_BoxLoops.H`, `CD_ParticleLoops.H`) keep correct nested indentation after conversion
- [x] `python3 -m sphinx -W --keep-going -b html source build/html` in `Docs/Sphinx`: 0 warnings (was 46), matching a from-scratch `main` build
- [x] Word-sequence content-preservation check confirms all ~100 restored `//` comment blocks (commit 5) are byte-identical in content to `main`
- [x] Repeated `clang-format` runs across the whole tree converge to a stable fixed point (zero further changes)
- [ ] CI: `Linux-GNU` / `Linux-oneAPI` 2-D compile-only build


rmrsk:
@claude review

rmrsk:
Addressed the \`ReflowComments\` / hanging-indent feedback in 3a82838e.

Turned out to be a bigger issue than the one example suggested — a systematic scan found ~100 more \`//\` comment blocks across ~35 files using the same "vary the spacing after \`//\` to hand-align continuation text" convention (not just \`TLDR:\`-style prose; also ASCII-art diagrams and numbered/lettered lists, e.g. \`CD_EBHelmholtzOp.cpp\`'s discretization sketch + procedure list). `ReflowComments` mishandles all of them the same way: it groups comment lines into "paragraphs" by the whitespace after \`//\`, so any block that intentionally varies that spacing gets split into independently-reflowed fragments — sometimes just an awkward line break, sometimes (for lists/diagrams) something that would be actively wrong if merged into flowing prose.

Fix: restored each affected block to its exact original text (re-indented to the current post-reformat indentation) and wrapped it in \`// clang-format off\` / \`// clang-format on\`, rather than trying to auto-flatten into reflowable prose — several of these are lists/diagrams where that would have been wrong, and there's no single transform that's correct for all of them. Verified content-preservation the same way as the literalinclude fix (word-sequence matching against \`main\`), and confirmed clang-format now reaches a stable fixed point on the whole tree (repeated runs produce zero further changes — a few unrelated lines needed one more pass than the original two-pass pipeline gave them). \`pre-commit run --all-files\` and a full \`sphinx-build -W\` both still pass clean.

PR #674, opened 2026-08-01
Move internal ParticleManagement helpers into a detail namespace

rmrsk:
# Summary

Relocate the three internal-only helpers in `ParticleManagement` into a `ParticleManagement::detail` namespace, mirroring the split already used for the nearest-neighbor merge. This is a pure refactor with no behavioral change.

### Background

The `ParticleManagement` API exposes a set of public factory/wrapper functions (`make*Merger`, `partitionParticleWeights`). Three helpers live behind those wrappers and are the algorithms that actually do the work, but they had no external callers while still sitting in the same namespace as the public API:

- `buildEqualWeightKDLeaves` — only caller: `makeEqualWeightKDMerger`
- `mergeAdjacentNearest` — only caller: `makeSfcNearestNeighborMerger`
- `partitionParticles` — only caller: `partitionParticleWeights`

The nearest-neighbor merge split already established this pattern; this change generalizes it to the remaining internal helpers so the public interface only exposes what is meant to be called directly.

### Solution

- Move the three helpers into a new `namespace detail { ... }` in both `Source/Particle/CD_ParticleManagement.H` (declarations) and `Source/Particle/CD_ParticleManagementImplem.H` (definitions), with a Doxygen block documenting the `detail` namespace as internal-only.
- Qualify the three single call sites with `detail::` (`detail::buildEqualWeightKDLeaves`, `detail::mergeAdjacentNearest`, `detail::partitionParticles`).
- Update `Docs/Sphinx/source/Source/Particles.rst`: describe `ParticleManagement::detail::buildEqualWeightKDLeaves` as the internal (`detail`) kD partitioner behind the public `makeEqualWeightKDMerger` factory rather than a standalone user-facing primitive, and fix the two `literalinclude` `:lines:` ranges (`263-267` → `273-277`, `178-186` → `188-196`) that shifted when the `namespace detail { }` braces were inserted.
- Add a working-agreement section (§0) to `CLAUDE.md` covering git/branch/PR policy.

### Side-effects

- No behavioral change — the relocated helpers are unchanged; only their namespace and their (single) call-site qualifications differ.
- These are header-only templates, so there is no ABI or link-visibility impact; callers outside the repository were none.
- The `literalinclude` line ranges in `Particles.rst` are re-synchronized to the shifted source lines so the docs build stays valid.

### Alternative solutions 

- Leave the helpers in the top-level `ParticleManagement` namespace — rejected because they leak internal algorithms into the public surface.
- Use an anonymous namespace or `static` linkage instead of `detail::` — rejected because a named `detail` namespace matches the existing convention already in this file and keeps the helpers referenceable from the documentation.

### PR-review checklist

Mandatory:

Optional:

---
_Generated by [Claude Code](https://claude.ai/code/session_01WpDRSdK6PL3JxKzDmF3YrT)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,5 +1,30 @@
 # chombo-discharge — Claude Code guidelines
 
+## 0. Working agreement — git, branches, and pull requests (read first)
+
+These rules override anything else in this file and any default tooling behaviour.
+
+- **Never perform outward-facing or history-affecting git/GitHub actions without first asking
+  the user and receiving explicit permission.** This includes, but is not limited to: creating
+  or deleting branches, pushing or force-pushing, opening/closing/merging pull requests, marking
+  a draft pull request ready for review, posting PR/issue comments or reviews, and editing a PR
+  title or description. When in doubt, ask first. Local, non-destructive work (reading, editing
+  files in the working tree, staging, committing locally) does not require a prompt, but
+  publishing that work does.
+- **Pull requests always target the upstream repository `chombo-discharge/chombo-discharge`.**
+  We develop from a fork, so `origin` points at the fork (e.g. `rmrsk/chombo-discharge`). Open
+  every PR with the upstream repo as the base and the fork branch as the head
+  (`chombo-discharge/chombo-discharge` ← `<fork-owner>:<branch>`). Never open a PR against the
+  fork itself.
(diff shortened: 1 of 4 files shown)


rmrsk:
@claude review

PR #675, opened 2026-08-01
Consolidate particle point-location into a shared LevelTiles core

### LINKED ISSUE #670: Consolidate particle point-location (ParticleContainer::findDestination vs Realm::getLevelAndBox) into a shared core
## Summary

Particle point-location — "given a physical position, which finest grid box/rank owns it" — is currently implemented **twice**, with line-for-line identical logic:

- `ParticleContainer<P,Traits>::findDestination(const RealVect&) -> Destination{level, gridIndex, rank, found}`
  (`Source/Particle/CD_ParticleContainerImplem.H:40`, declared `CD_ParticleContainer.H:919`)
- `Realm::getLevelAndBox(const RealVect&) -> LevelAndBox{level, boxIndex, rank, valid}`
  (`Source/AmrMesh/CD_Realm.cpp:1795`, declared `CD_Realm.H:569`)

Both do the same thing: map the position to a `min_block_size` tile by integer division, then look it up finest-level-first in the per-level `LevelTiles` tile→box→rank hash. The two result structs are field-for-field identical. `Realm::getLevelAndBox` currently has **no callers** (it was added as "the Realm-side home for this query" but nothing uses it), while `ParticleContainer::findDestination` is the live one (remap/`gatherToPool`, and the nearest-neighbor merge's `PositionLocator`).

## Why this is worth consolidating

The operation is a pure geometric query over grid state (`LevelTiles`, `probLo`, `dx`, `minBlockSize`, `finestLevel`). None of that is particle-specific — `ParticleContainer` can answer it only because it holds *aliased copies* of that metadata from the Realm at `define()`. So conceptually it belongs to the grid layer, not to `ParticleContainer`; the container is a consumer. Two independent copies can (and eventually will) drift.

## Proposal

Extract **one** shared, inline core and have both call sites delegate to it. Two constraints shape *where* the core lives:

1. **Hot-path inlinability.** `findDestination` is called *per particle* in `gatherToPool`/remap, concurrently under OpenMP. It is currently a header-inlined read of local members. The shared core must stay in a header so this loop keeps inlining — routing it through a compiled `Realm` method would de-inline the per-particle path.
2. **`ParticleContainer` is deliberately `Realm`-agnostic.** It includes `CD_LevelTiles.H` but not `CD_Realm.H` (it takes the tiles as an alias, not a `Realm&`; the merge header is likewise kept AmrMesh-agnostic). Making `findDestination` call `Realm::…` would invert that and pull `Realm` into the low-level container.

Both point to hosting the core **one level below `Realm`, on/beside `LevelTiles`** (which both `Realm` and `ParticleContainer` already include), e.g.:

```cpp
// CD_LevelTiles.H (inline free function or static)
LevelAndBox findDestination(const RealVect& a_pos,
                            const RealVect& a_probLo,
                            const Vector<Real>& a_dx,          // isotropic per level, matching AmrMesh::getDx()
                            int a_minBlockSize,
                            const Vector<RefCountedPtr<LevelTiles>>& a_levelTiles,
                            int a_finestLevel);
```

Then:
- `Realm::getLevelAndBox` becomes a one-line wrapper (or rename to `Realm::findDestination` — free to do, it has no callers).
- `ParticleContainer::findDestination` becomes a one-line wrapper passing its aliased members — still inline, still no `Realm` dependency.
- Unify `Destination` / `LevelAndBox` into a single struct next to the core.
- The merge builds its `PositionLocator` from this core / `Realm` directly (it already holds `AmrMesh&`), dropping its reliance on `ParticleContainer::findDestination`.

Hosting the static on `Realm` (the originally-floated option) also improves on today, but forces a choice: either `ParticleContainer` gains a `Realm` include on the per-particle path, or it keeps its own copy — i.e. it doesn't fully kill the duplication. The `LevelTiles` home avoids that.

Low risk overall, since the Realm-side copy has no callers yet, and the change is a pure refactor with no behavioral change.

## Related idea: warm-start remapping (movers only)

Motivation noted for the record. After a CFL-limited time step, a particle almost always stays in its box or crosses into an *adjacent* box, so most of remap should be near-free.

Two clarifications on the current state, since the framing assumed `findDestination` is a "cold-start that scans all patches":

- `findDestination` is already **O(1)** — a tile-index computation plus an `unordered_map` lookup per level, **not** a linear scan over all grid patches. So the "avoid scanning every patch" win is already realized.
- Same-box **stayers are already short-circuited**: remap's mask fast path leaves a particle in place when it is still inside its box and not under a finer level, and only *movers* are pooled and sent through `findDestination`.

So the remaining opportunity the warm-start targets is narrower: for **movers** (particles that left their box), check the handful of neighbor boxes adjacent to the box the particle came from before falling back to the tile-hash. Trade-offs to weigh:

- Upside: skips the tile computation + per-level hashing for the common "moved to an adjacent box" case.
- Downside: up to `3^D - 1` neighbor `Box::contains` checks (26 in 3-D) can cost more than a single hash lookup; it also needs local box-adjacency info, and cross-rank/cross-level neighbors still need the fallback.

Given `findDestination` is already O(1) and stayers already bypass it, this is a **measure-first** optimization — worth prototyping against a representative advection case to see whether the mover fast path actually beats the hash. It's cleaner to pursue on top of the consolidation above (single point-location core to instrument and to add a fast path in front of).

## Files

- `Source/Particle/CD_ParticleContainer.H` / `...Implem.H` — `findDestination`, `Destination`, remap/`gatherToPool`
- `Source/AmrMesh/CD_Realm.H` / `.cpp` — `getLevelAndBox`, `LevelAndBox`
- `Source/AmrMesh/CD_LevelTiles.H` — proposed home for the shared core
- `Source/Particle/CD_NearestNeighborParticleMergeImplem.H` — the merge's `PositionLocator`

rmrsk:
# Summary

Particle point-location -- "given a physical position, which finest grid box/rank owns it" -- was implemented **twice** with line-for-line identical logic and field-for-field identical result structs:

- `ParticleContainer::findDestination` (the per-particle remap hot path)
- `Realm::getLevelAndBox` (no callers)

This extracts **one** shared, header-inlined core, `LevelTiles::findDestination`, and has both sites delegate to it. Pure refactor, no behavioral change. Closes #670.

### Background

Both functions mapped a position to a `min_block_size` tile by integer division, then looked it up finest-level-first in the per-level `LevelTiles` tile->box hash, and returned identical `{level, index, rank, valid}` structs. The operation is a pure geometric query over grid state (`LevelTiles`, `probLo`, `dx`, `minBlockSize`, `finestLevel`) -- none of it is particle-specific; `ParticleContainer` can answer it only because it holds aliased copies of that metadata from the `Realm`. Two independent copies can drift.

### Solution

- Add an inline `LevelTiles::findDestination(...)` core plus a unified `LevelTiles::LevelAndBox` result struct, hosted on `LevelTiles` -- which both `Realm` and `ParticleContainer` already include. This keeps the core header-inlinable for the hot path and does **not** pull `Realm` into the low-level container (`ParticleContainer` stays `Realm`-agnostic).
- `ParticleContainer::findDestination` becomes a one-line wrapper passing its aliased members, so the per-particle path keeps inlining with no conversion.
- `Realm::getLevelAndBox` becomes a thin wrapper that delegates to the core.
- The two result structs (`ParticleContainer::Destination`, `Realm::LevelAndBox`) are unified into `LevelTiles::LevelAndBox`; the nearest-neighbor merge's `PositionLocator` uses the shared type.
- `dx` representation: the core takes `Vector<RealVect>` (`ParticleContainer`'s native form) so the hot path passes `m_dx` with zero conversion and no new member. `Realm` (no callers, cold) adapts its isotropic scalar `m_dx` once before delegating. The alternative scalar `Vector<Real>` signature would have rippled into `ParticleContainer::getDx()` (public, returns `Vector<RealVect>`, consumed by `ParticleOps`).
- The merge keeps routing through `ParticleContainer::findDestination`, which is now itself a thin delegate to the shared core -- so the single-source-of-truth guarantee holds without adding new `AmrMesh`/`Realm` coupling to the merge.

### Side-effects

- No behavioral change; the relocated arithmetic is unchanged.
- The core is header-only/inline, preserving the per-particle hot path's inlinability.
- `Realm::getLevelAndBox` now builds a small per-level `Vector<RealVect>` from its scalar `dx` before delegating -- off any hot path (the method has no callers).
- No `literalinclude` line ranges were affected (the edits in `CD_ParticleContainer.H` sit past its last referenced range); the `Realm.rst` point-query example was updated to the unified type/field names.

### Alternative solutions 

- **Host the core on `Realm`** (the originally-floated option): forces `ParticleContainer` to either gain a `Realm` include on the per-particle path or keep its own copy -- i.e. it does not fully kill the duplication. Hosting on `LevelTiles` avoids that.
- **Scalar `Vector<Real>` dx in the core signature:** cleaner abstraction, but would require either a redundant scalar member on `ParticleContainer` or changing `m_dx`'s type and thus the public `getDx()` contract. Deferred in favor of the zero-overhead `Vector<RealVect>` form.
- **Rewire the merge to build its `PositionLocator` from `Realm`/the core directly:** deferred -- `findDestination` is now a thin delegate, so this yields no functional change while adding coupling. Straightforward follow-up if desired.

### PR-review checklist

Mandatory:

Optional:

---
_Generated by [Claude Code](https://claude.ai/code/session_01WpDRSdK6PL3JxKzDmF3YrT)_

### DIFF
--- a/Docs/Sphinx/source/Source/Realm.rst
+++ b/Docs/Sphinx/source/Source/Realm.rst
@@ -47,11 +47,11 @@ The lookup is exposed through ``Realm::getLevelAndBox``, which returns the fines
 
 .. code-block:: c++
 
-   const Realm::LevelAndBox result = realm.getLevelAndBox(position);
-   // result.level    -- AMR level containing the point (-1 if none)
-   // result.boxIndex -- global grid-box index on that level
-   // result.rank     -- MPI rank owning that box
-   // result.valid    -- false if no box covers the point (i.e. off-domain)
+   const LevelTiles::LevelAndBox result = realm.getLevelAndBox(position);
+   // result.level     -- AMR level containing the point (-1 if none)
+   // result.gridIndex -- global grid-box index on that level
+   // result.rank      -- MPI rank owning that box
+   // result.valid     -- false if no box covers the point (i.e. off-domain)
 
 This is the same mapping the particle infrastructure uses to bin particles during ``remap()`` (see :ref:`Chap:Particles`).
 ``ParticleContainer`` aliases the realm's tile hash grid rather than building its own, so the ``Realm`` is the single source of truth for point-to-patch ownership.
--- a/Source/AmrMesh/CD_LevelTiles.H
+++ b/Source/AmrMesh/CD_LevelTiles.H
@@ -14,12 +14,17 @@
 #define CD_LEVELTILES_H
 
 // Std includes
+#include <cmath>
 #include <cstddef>
 #include <unordered_map>
 #include <vector>
 
 // Chombo includes
 #include <IntVect.H>
+#include <RealVect.H>
+#include <Vector.H>
+#include <RefCountedPtr.H>
+#include <SPMD.H>
 #include <DisjointBoxLayout.H>
 #include <DataIndex.H>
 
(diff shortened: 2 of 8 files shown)


rmrsk:
@claude review

Issue #665, opened 2026-07-06
BVH particle bucket merge


rmrsk:
It might be possible to bucket-merge large clusters of particle with a patch-local BVH merge, including ghost cells. This would prevent cell aliasing aking to the nearest-neighbor merge, and wouldn't require building the tree all the way down either.

The biggest issue with this merge type would be to the implementation of tie-breakers across ranks, which is non-trivial also for the nearest-neighbor merge. It might be important to ensure that the BVH volume is smaller than a grid cell size to avoid spurious merges. We do the same in all other types of merge, so this step will be simple.

rmrsk:
## Design note: patch-wide kD merge with AABB leaf-proposals for the boundary

Coming out of profiling the current `nn_pair` merge (`mergeNearestNeighborsRound`), where per-round cost is dominated roughly equally by the gather, the `allNearestNeighbors` spatial-index build, and the pairwise candidate search — and the whole thing is repeated over several drain rounds. This sketches a BVH/kD-per-patch alternative that could be single-pass.

### The idea

Instead of per-cell nearest-neighbor pairs merged over multiple rounds, build a kD/BVH over each **patch**'s particles (a variant of `equal_weight_kd` applied to a whole patch rather than a single cell) and descend only far enough to form **merge-group leaves** (leaf weight ≤ target). Merge each leaf in one shot.

Why it's attractive: it directly removes the three dominant costs — no `allNearestNeighbors` graph, no per-particle candidate search, and **no drain-round loop** (nn_pair needs ~log(surplus) rounds; a kD descent reaches the target in one pass). For fully-local leaves there's also no cross-rank coordination at all.

### The boundary problem

A kD partition is a **global** operation over the patch's point set. Neighboring patches partition **different** sets (each sees its own particles plus only its own ghost halo), so their boundary leaves do **not** agree: patch A may group {a1, a2, b1_ghost} while patch B groups {b1, b2, a1_ghost}. With `nn_pair` the atomic unit is a *pair* and "nearest neighbor" is a symmetric, identically-computable relation, so propose/judge/verdict converges to one answer. A multi-particle *group* has no canonical cross-patch definition, so a tie-break has nothing to select between — the groupings themselves conflict. Widening the halo doesn't fix it; the split path still depends on the full (differing) sets.

### Proposed boundary handling: AABB leaf-proposals + a regenerating judge

[...]

PR #678, opened 2026-08-04
Add kd_carve and kd_patch particle merges (#665)

rmrsk:
# Summary

Adds a distributed, MPI-safe whole-patch super-particle merge built on a kd-tree partition under a per-cell leaf quota, addressing #665. Two variants are selectable: `kd_carve`, which arbitrates patch boundaries with a non-iterative "z-buffer carve", and `kd_patch`, which is purely patch-local.

### Background

The existing distributed merge, `nn_pair_tree`, matches particles pairwise off a k-NN graph and drains through repeated rounds. Issue #665 asked for an alternative built on a whole-patch spatial partition instead, with the patch/rank boundary resolved deterministically rather than by iteration.

The purely local merges (`equal_weight_kd`, `reinitialize`, `nn_sfc`) are not usable where merging must cross a cell face: confining a merge to one cell locks the pattern to the mesh and aliases.

### Solution

One kd-tree per patch over the patch's particles, split by position at the longest axis and never snapped to the grid, so a leaf may straddle a cell face.

How far splitting goes is set by a **live per-cell leaf quota**. A leaf collapses to exactly one super-particle at its weighted centroid, so the number of leaves centred in a cell is that cell's post-merge population; that count is tracked while splitting and a split that would push a cell past `ppc` is refused. Slots go to the heaviest candidate leaf first, since the quota is claimed first-come-first-served and a count ordering strands heavy leaves whole.

The split plane is the count median while a node is wider than `kd_weight_split_scale` cell widths and the weight median once narrower: above that scale a split apportions leaves between cells, below it the weight median drives super-particles toward equal weight.

The quota constrains only leaves that can actually merge. A leaf wider than one cell is vetoed and leaves its members behind individually, so it costs its cell its full member count rather than one particle, and the quota does not block splitting it further.

Two methods share this build:

* **`kd_carve`** — fills a width-1 particle ghost halo. Leaves clear of any patch or level interface commit immediately with no communication; the rest become candidate boxes keyed by `(AABB volume, anchor particle id)` and are resolved in three fixed exchange rounds (claim, verdict, commit), with no drain loop.
* **`kd_patch`** — fills no ghosts. Every leaf merges regardless of boundary exposure and no particle is ever contested, so each patch reduces only the particles it owns. Communication-free, at the cost of no coordination across patch boundaries.

Measured for `kd_carve` against `nn_pair_tree` on `BrownianWalker/KDCarve` at np=4, aggregated over all ranks and timesteps:

| | `kd_carve` | `nn_pair_tree` |
|---|---|---|
| mean in-cell weight CV | 0.116 | 0.184 |
| worst in-cell weight ratio | 228x | 1132x |
| cells over target | 0.74% | 0.90% |
| worst single cell | +12 | +1119 |
| max cell population (`ppc`=32) | 44 | 1151 |
| wall clock | 10.5 s | 25.0 s |

### Side-effects

- New `merge_algorithm` values `kd_carve` and `kd_patch`.
- New mandatory input `ItoSolver.kd_weight_split_scale` (default 1.0), added to every `.inputs` file carrying `ItoSolver` options.
- New regression tests `BrownianWalker/KDCarve` and `BrownianWalker/KDPatch`, with 2d/3d blocks in `BrownianWalker.ini`.
- No change to any existing merge algorithm.

Known limitation: roughly 0.74% of cells finish above target, worst +12. A cell can exceed `ppc` when a leaf is both too wide to merge and in a cell already at quota, since vetoing it leaves all its members behind while splitting it exceeds the leaf count. This is inherent to satisfying the quota and the extent bound at once. It is not a boundary effect — the rate is the same at np=1 and np=4, and the same in the patch interior as at patch edges.

### Alternative solutions 

- Driving the recursion by leaf extent rather than the quota: splits on geometry regardless of how many particles remain, shredding crowded cells into unmergeable singletons.
- Per-node budget rules (stop when a node is worth <= 1 or < 1.5 leaves): a per-cell quota cannot be expressed as a test a node applies to itself, and both thresholds either split every mergeable pair or keep every one.
- Grid-aligned grouping (sort by cell, split within): hits `ppc` exactly but confines leaves to a cell and aliases.
- Booking the quota in predicted particles rather than leaf slots: the root starts far outside the budget, so an absolute test then refuses nearly every split.
- Exposing `kd_patch` as a boolean on `kd_carve` rather than its own method: it changes the contract (no ghosts, all leaves eligible) rather than tuning behaviour, and "carve with the carve off" is a contradiction.
- A post-merge merge-down/rebalance pass in `splitAndRebuildFromMergeContainer`: rejected as it would alter other merge algorithms sharing that path.

### PR-review checklist

Mandatory:

Optional:


rmrsk:
@claude review

rmrsk:
@claude review

rmrsk:
@claude review

rmrsk:
@claude review

PR #681, opened 2026-08-08
Add kd_skin_nn: whole-patch kd-tree merge with a nearest-neighbour skin

### LINKED ISSUE #680: Remaining particle mergers
There are two more particle mergers I'd like to test:

1. Bucket BVH merge with NN skin treatment
2. An non-ghosted BVH merge akin to the existing bucket merge, but disregarding neighbor merges (PS: Will create patch artifacts)

rmrsk:
# Summary

Adds `kd_skin_nn`, a third whole-patch kd-tree merge alongside `kd_carve` and `kd_patch`. It keeps
the kd tree build and the live per-cell quota, but hands the contested boundary region to the
existing nearest-neighbour pair merge instead of `kd_carve`'s bespoke arbitration.

Closes #680 

### Background

`kd_carve` (#678, issue #665) reduces each patch's particles with a whole-patch kd tree under a live
per-cell quota, and resolves particles near a patch boundary with a three-round "z-buffer carve": a
`(volume, anchor-id)` key protocol and a claim/verdict/commit handshake whose entire purpose is to
stop two patches merging the same particle twice. That tier is the expensive and intricate part of
the algorithm.

It is also conservative in a second way. `kd_carve` commits a leaf locally only when no member's
cell is boundary-exposed **and** the leaf holds no ghost, so leaves that are exposed but not actually
contested are sent to arbitration unnecessarily.

### Solution

The particles are split across two containers.

`ParticleManagement::mergeKDInterior()` commits every leaf that holds **no ghost** into an interior
container and removes its members from the working one. What it could not commit — leaves holding a
ghost, and leaves too wide to collapse — stays behind as the *skin*, and is drained by
`ParticleManagement::mergeNearestNeighborsOneCell()`, whose search radius is structurally fixed at
Chebyshev cell distance 1 and therefore matches the width-1 ghost halo and the one-cell leaf-extent
bound exactly. The two containers are then concatenated per patch and handed to the existing
split-and-rebuild.

"Holds no ghost" is the entire safety condition, and it costs no communication to evaluate. Every
member of such a leaf is physically resident in this patch; any other patch that draws one of those
members into a leaf of its own necessarily sees it as a ghost, which disqualifies that leaf there by
this same test. No two patches can commit the same particle, so no weight is counted twice.
`kd_carve`'s additional `!anyExposed` requirement is arbitration bookkeeping rather than a safety
condition, and dropping it is what makes the skin small.

The two tiers share **one** per-cell target. Every interior super-particle reserves a slot in the
cell it landed in, so the skin tier drains each cell to what is *left* of the target rather than to
the whole of it. This required generalising the nearest-neighbour merge's crowding target from a
bare `int` to a new `ParticleManagement::NNCellBudget`, which is implicitly constructible from an
`int` so that every existing caller and default argument is unchanged.

Two details of the reservation have to match how `gatherMergeParticles()` builds its own live
per-cell count, or the budget is measured against the wrong denominator at exactly the seams where it
matters: reservations are booked once per **distinct particle id** (an interior super-particle is
resident in its own patch and simultaneously a ghost in every neighbouring patch the rank owns), and
a ghost is booked against the level it came **from**, with that level's `dx`.

### Side-effects

- **`nn_pair_*` results shift at round-off.** The merge *decisions* are unchanged — the integer
  particles-per-cell field is identical to `main` step for step — but the deposited density differs
  by about 1 ULP (measured 1.16e-10 on 9.3e5, i.e. 1.2e-16 relative) on the first step, because
  particle ordering and hence the deposition summation order changes. Over a long run that round-off
  amplifies chaotically, so **`nn_pair_*` benchmark files will change** even though the algorithm
  does not. Confirmed to originate in the merge path: with `merge_algorithm=none` the two builds are
  byte-identical.
- `ParticleMergeMethod` gains a `KdSkinNn` enumerator, so `External`'s underlying value shifts.
- No new ParmParse keys. The skin tier reuses `nn_pair_iterate`, `nn_pair_fallback` and
  `nn_pair_max_rounds`; `nn_pair_max_cell_dist` does not apply, since the radius is fixed at 1.
  `kd_split_weight_leaf_dx` governs the shared tree build and so applies here too.
- A cell holding interior results plus a single leftover skin particle finishes one particle above
  target, because a skin particle and an interior super-particle live in different containers and can
  never merge with each other. This is the same class of residual as `kd_carve`'s ~0.74% and is not a
  conservation error.

### Alternative solutions 

- **Keep the carve tier.** `kd_carve` remains available and unchanged; this is an additional
  algorithm, not a replacement.
- **`nn_pair_tree`/`nn_pair_hash` for the skin.** Rejected: both permit a merge radius wider than one
  cell, which would require filling the skin container's ghosts at that wider width and registering a
  matching mask. `nn_pair_onecell`'s fixed radius already agrees with the kd tier's bounds.
- **Leave wide ghost-free leaves unmerged**, as `kd_carve`'s extent veto does. Rejected: routing them
  to the skin instead lets the nearest-neighbour tier reduce them, so those cells still reach the
  target.
- **Separate `kd_skin_nn_*` tunables.** Rejected as three more mandatory inputs with semantics
  identical to the existing keys.
- **Per-cell budget by other means.** Passing a plain scalar would leave boundary cells
  under-merged, since the skin tier would drain each cell to the full target on top of the interior
  results already there.

### PR-review checklist

Mandatory:

Optional:


rmrsk:
@claude review

PR #684, opened 2026-08-09
Use mesh containers for the particle mergers' per-cell data, and remove duplicated types

### LINKED ISSUE #682: Particle mergers: per-cell bookkeeping duplicates EBAMRCellData/LevelData functionality
The distributed particle mergers (`nn_pair_*`, `kd_carve`, `kd_patch`) carry a fair amount of
bespoke per-cell bookkeeping that duplicates functionality already provided, and tested at scale, by
`EBAMRCellData` / `LevelData` / `BaseFab` — including coarsening, ghost filling, interpolation, and
accumulation back out of ghost cells via `Copier` + `LDaddOp`.

This issue inventories what is actually stored per cell, so the scaffolding can be assessed against
the mesh-data types rather than grown further. It is an audit, not a proposed patch.

Line references are against `main` unless noted.

### Patch-local scratch — the analogue is a `BaseFab`/`FArrayBox` over the patch box

| Store | Current type | Mutated | Notes |
|---|---|---|---|
| `exposureCache`, `mergeKDCarve` (`CD_KDParticleMergeImplem.H:871`) | `std::vector<signed char>`, dense over the patch box | memoized | **Already a FAB, hand-rolled.** The linear index is computed manually: `cellIdx += (a_iv[dir] - patchLo[dir]) * stride`. This is `BaseFab<signed char>` over `dbl[din]` with extra steps. |
| `used`, `buildKDQuotaLeaves` (`CD_KDParticleMergeImplem.H:535`) | `std::map<IntVect, int, decltype(cellLess)>` | **continuously** — this is the live per-cell leaf quota | Would need the box grown by one ghost cell: ghosts are leaf members, so a leaf centroid can land outside the patch's own box. |
| `KDCellHistogram::m_counts` (`CD_KDParticleMergeImplem.H:240`) | sorted `std::vector<std::pair<IntVect,int>>` + binary search | no | Same ghost caveat — ghost particles contribute to the counts. |
| `cellCount`, `nnMergeAnyCellCrowded` (`CD_NearestNeighborParticleMergeImplem.H:295`) | `std::unordered_map<NNWalkCell, int, LevelTiles::TileHasher>` | no | Rebuilt per patch and discarded. It is a per-patch restriction of `a_liveCellCount` below, so it is arguably deletable outright rather than replaced. |

### Hierarchy-wide — genuine `EBAMRCellData` candidates

| Store | Current type | Mutated | Notes |
|---|---|---|---|
| `a_liveCellCount` (6 declarations in `CD_NearestNeighborParticleMergeImplem.H`) | `std::unordered_map<NNCellKey, int, NNCellKeyHasher>` | **yes, in inner loops, including on ghost cells** | Decremented as each merge commits, on both local and ghost cells. This is the case needing the `LDaddOp` / reverse-copier accumulate pattern rather than a plain `exchange()`. |
| `NNCellBudget::m_reserved` (introduced in #681, not in `main`) | `std::unordered_map<NNCellKey, int, NNCellKeyHasher>` | no, read-only after construction | Spans levels. Currently populated by an *extra particle halo exchange* of a second container purely to learn remote reservations; a scalar `copyTo`/`exchange` would move one value per ghost cell instead of a full particle record per ghost particle. |

## Per-cell keyed, but not mesh data

`NNCellSpatialIndex` (`unordered_map<NNWalkCell, shared_ptr<NNCellCloud>, LevelTiles::TileHasher>`)
is keyed by cell but its value is a `PointCloudBVH` pointer, so it cannot live in a FAB. Only its
keys could come from a cell iterator.

## Named for cells but actually per-particle

`NNCellMeta` (`CD_NearestNeighborParticleMerge.H:614`) holds `globalID`, `isLocal` and `level` per
BVH primitive — it is per-particle metadata. The name suggests otherwise. `KDLeaf` is per-leaf (index
range plus bounding box), not per-cell.

## Cascade

`NNCellKey` (`CD_NearestNeighborParticleMerge.H:375`), `NNCellKeyHasher`, `NNWalkCell`, and the
`kdIntVectLess` / `cellLess` comparators exist *solely* to key the maps above safely. The comparators
exist specifically because `IntVect::operator<` is a component-wise partial order rather than a
strict weak ordering, so using it as a `std::map`/`std::set` comparator silently aliases distinct
cells. Moving the stores to cell-indexed data removes three aliases and two comparators with them.

## Unrelated duplication found while auditing

`NNParticleLocation` (`CD_NearestNeighborParticleMerge.H:1010`) and its `PositionLocator` callback
are a field-for-field duplicate of `LevelTiles::LevelAndBox`, which
`ParticleContainer::findDestination()` already returns. The code repackages one into the other:

```cpp
// CD_NearestNeighborParticleMergeImplem.H:1517
return NNParticleLocation{dst.valid, dst.level, dst.gridIndex, dst.rank};
```

A struct plus a `std::function` indirection that exist only to rename four fields.

## Not candidates

For completeness, so these are not re-litigated: the MPI wire formats (`KDClaim`, `KDVerdict`,
`KDCommit`, `NNMergeProposal`, `NNMergeVerdict`, `NNMergeResult`) must remain flat POD types, and the
packed particle types (`KDParticle`, `NNMergeParticle`) carry per-particle state that the existing
`NonCommParticle<M,N>` does not provide — it has no id, owner rank or ghost flag. `KDParticle` and
`NNMergeParticle` are, however, near-duplicates of each other, differing only in `isGhost` vs
`level`.

## Caveats for anyone acting on this

- **Ghost cells are required, not optional**, for the two patch-local KD stores: ghosts participate
  in both the quota and the histogram, and leaf centroids can fall outside the patch's own box.
- **`cellKeyOf` is deliberately unclamped** — "not clamped to any patch's box, so a ghost buckets
  into its true origin cell". A `BaseFab`/`LevelData` is defined over a `Box`, so any conversion
  needs an argument that every consulted key lands inside an allocated (ghosted) box. That holds at
  ghost width 1 but becomes an invariant to maintain rather than something the type enforces.
- **Multi-valued cut cells**: none of the current stores distinguish multiple VoFs in a cell — a bare
  `IntVect` key conflates them. `EBAMRCellData` represents them properly, so this would be a
  behaviour change rather than a pure refactor. Probably a desirable one, but it should be a
  deliberate decision.
- **Nothing here is measured.** No profiling or scaling study backs the performance reasoning; the
  largest run behind these observations was 8 ranks on a single machine.

## Explicitly out of scope

Merging the interior and skin particle containers in `kd_skin_nn` (#681) was considered and
**rejected**: the nearest-neighbour tier builds a `PointCloudBVH` per occupied cell over every
particle in a patch once any cell there is crowded, so a single container would make the index build
O(volume) instead of O(surface area). The two-container split is a performance property, not
bookkeeping.

rmrsk:
# Summary

Replaces the particle mergers' hand-rolled per-cell bookkeeping with the mesh containers the codebase
already provides, removes ten types, and folds two near-duplicate merge bodies into one. Behaviour is
unchanged throughout, and every step is verified byte-identical rather than argued to be equivalent.

This is the tractable part of #682. Two items from that audit were attempted or costed and
deliberately not done — see *Alternative solutions*.

Closes #682 

### Background

#682 audited the distributed mergers and found six per-cell stores duplicating functionality
`EBAMRCellData`/`LevelData`/`BaseFab` already provide, tested at scale. Beyond the duplication, the
hand-rolled versions left indexing invariants implicit: a map silently accepts any key, which is how
two budget bugs got into #681 without erroring.

A broader concern was raised while reviewing that audit — whether the mergers had accumulated enough
scaffolding to be hard to maintain. Measurement supported it: 32 types across the two file pairs,
several duplicative, five of them used by exactly one function, and two merge bodies that were 90%
the same function.

### Solution

**Per-cell data now uses mesh containers.** The kd histogram and live leaf quota are `EBAMRFAB`
allocated once per merge on the realm with one ghost cell, accessed per patch as `FArrayBox&`. The
exposure cache became a `BaseFab`, replacing linear indexing computed by hand
(`cellIdx += (a_iv[dir] - patchLo[dir]) * stride`). Containment is asserted at each site, so the
indexing invariant is enforced rather than assumed. `AmrMesh` gains
`allocate(EBAMRFAB&, realm, nComp, ghost)` — the type already existed, only the overload was missing
while its sibling `EBAMRBool` had one.

**Types removed**, all provably redundant rather than merely tidied:

- `NNParticleLocation` duplicated `LevelTiles::LevelAndBox` field for field — its own doc said so —
  and the code repackaged one into the other. It now uses `LevelAndBox` directly.
- `cellLess` existed only because `IntVect::operator<` is a component-wise partial order that an
  ordered container would use to silently pool unrelated cells. The map it guarded is never iterated,
  only point-queried, so the ordering bought nothing.
- `kdIntVectLess` went with `KDCellHistogram`, having had no other user.
- `KDCellHistogram` became a function filling a caller-owned holder.
- `KDBoxKey`, `KDMember`, `KDClaim`, `KDVerdict` and `KDCommit` moved into `mergeKDCarve`'s body.
  Each is used by exactly one function yet sat in the public header with a full documentation block.
- `KDParticle` and `NNMergeParticle` became one `MergeParticle<Packed>`. They differed by exactly one
  field — `isGhost` versus `level` — and were otherwise the same type documented and maintained
  twice. 153 references renamed.

**`mergeKDPatch` folded into `mergeKDInterior`.** The bodies were 90% identical, and the three
apparent differences reduce to one: `mergeKDInterior` commits only ghost-free leaves, and
`kd_patch`'s contract is that no ghost halo is filled, so every leaf is ghost-free and that test
admits them all. Passing the container as its own destination reproduces the rest. No flag was
needed; 237 lines became 38.

Net effect on the public surface: the kd header goes from 7 types to 1 and from 535 to 398 lines; the
nn header from 14 types to 12.

### Side-effects

- `AmrMesh` gains one overload.
- The carve protocol's wire layout is no longer in the generated Doxygen. That is intended — nothing
  outside `mergeKDCarve` could use it — but the layout is now discoverable only by reading that
  function.
- The three carve wire types are function-local types passed as template arguments to the exchange
  helper. Legal from C++11 on, and safe across ranks because every rank runs the same binary with the
  same layout.
- `MergeParticle` is appended at the END of `CD_ParticleManagement.H` deliberately: the `.rst`
  literalincludes cite absolute ranges at 198-206 and 283-287, and inserting above them would
  silently shift the rendered excerpts. The reason is recorded in the type's own docs so it is not
  tidied upward later.
- Nine `literalinclude` ranges in `MeshData.rst`/`Particles.rst` did need rebasing, because adding the
  `EBAMRFAB` overload shifted `CD_AmrMesh.H` by 14 lines. This is the silent kind of drift — ranges
  stay valid and Sphinx still exits 0, it just renders the wrong excerpt — so each was verified by
  comparing the rendered excerpt against what the original range captured, not by trusting the
  arithmetic.
- No behaviour change, so no benchmark files should move.

### Alternative solutions 

**Converting `a_liveCellCount` and `NNCellBudget` to mesh data was attempted and abandoned.** It is
listed in #682 and does not survive contact with the code. Those are rank-global, cross-level state
mutated inside the merge loop. Making them mesh data required patch-scoping the counts, which in turn
required a patch index on the pooled particle struct, a new field on an MPI wire struct so the judge
could learn a remote cell's count, and — fatally — introduced a cross-patch staleness hazard that does
not exist today: a cell visible to several patches would hold several independent copies of its
count, and committing a merge decrements only one. Not a conservation bug, since `consumedIDs` stays
rank-wide, but it permits over-merging a shared cell. Simple, exact and rank-global is the right shape
for that quantity; the map stays.

**Consolidating the two nearest-neighbour round drivers was costed and skipped.** `nn_pair_tree` and
`nn_pair_hash` already share one body via the `Cloud` template parameter; the remaining duplication is
whole-patch versus per-cell, at 82% similarity. Unlike the kd fold, those differences do not collapse
— the two hold different index types, built by different functions and queried by different functions.
Merging them needs either a new policy abstraction or three `if constexpr` branches inside a 200-line
orchestrator, which relocates complexity rather than removing it.

**`EBAMRCellData` rather than `EBAMRFAB`** was the original intent, but it requires a `phase` the
mergers do not have and should not need, and its EB machinery would be allocated only to be discarded
through `getFArrayBox()`. These are counts, and interpolating or coarsening a particle count is not a
defined operation. `EBAMRFAB` is phase-free and gives exactly the plain-scalar semantics wanted; the
switch is one line per site if a genuinely interpolatable field ever appears.

**Aliasing the two particle types** rather than renaming would have been a ten-line change, but it
leaves two names for one type and makes the kd header depend on the nn one.

### PR-review checklist

Mandatory:

Optional:


rmrsk:
@claude review

rmrsk:
@claude review

Issue #679, opened 2026-08-05
Point-to-point (neighbor-restricted) particle MPI exchange for ghost-fill, mergers, and remap


rmrsk:
## Summary

Several particle-related MPI exchanges use a full-communicator `MPI_Alltoall`/`MPI_Alltoallv` (looping over every rank in the job) where the actual traffic is sparse -- each rank only ever needs to talk to a bounded set of "neighbor" ranks. At large rank counts this is a real cost: `Alltoall`/`Alltoallv`, even with well-optimized collective algorithms, costs at best O(log P) synchronized rounds every rank must participate in, versus O(1) for point-to-point `Isend`/`Irecv` restricted to a fixed small neighbor set.

## Where this applies

1. **`ParticleContainer::fillGhostParticles()`** (`Source/Particle/CD_ParticleContainerImplem.H`, via `distributeFromPool()`) -- builds temporary read-only ghost copies using `ParticleGhostMask`/`AMRParticleGhostMask` (`Source/AmrMesh/CD_ParticleGhostMask.H`, `CD_Realm.cpp`'s `defineParticleGhostMaskSameLevel`/`FineToCoar`/`CoarToFine`). This is **bounded**: a rank's ghost-neighbor set can be derived with zero MPI (see prototype below), since ghost adjacency is symmetric and the masks already record every outgoing target.
2. **Particle mergers' own claim/verdict/commit-style exchanges** -- e.g. `bucket_tree_carve`'s `bucketTreeExchangeByRank` (`Source/Particle/CD_BucketTreeCarveParticleMergeImplem.H`), and likely the equivalent in `nn_pair_tree`/`nn_pair_hash`/`nn_pair_onecell` (`CD_NearestNeighborParticleMergeImplem.H`). Also bounded, for the same reason as (1) -- these only ever route to ranks a particle could plausibly have been gathered as a ghost from/to.
[...]

PR #687, opened 2026-08-11
Cache the mergers' regrid-invariant data on the realm

### LINKED ISSUE #685: Cache the mergers' regrid-invariant, non-templated data instead of re-deriving it per call
The particle mergers re-derive several quantities on every call that cannot change between regrids.
This issue lists them, and proposes the filter for deciding what may be cached centrally.

## The filter: non-templated invariants only

Only structures that are **(a) invariant between regrids** and **(b) not templated** belong in
`AmrMesh`/`Realm`.

The templating half matters as much as the invariance half. Most of the mergers' working state is
parameterised on the particle payload (`Packed`), the SoA type (`P`, `Traits`), or the point-cloud
backend (`Cloud`). `AmrMesh` cannot sensibly hold per-solver template instantiations, and those
structures are particle-dependent anyway, so they are excluded twice over. What is left is plain mesh
data and plain integers, which is exactly what the existing operators already cache.

Precedent: `EBAMRParticleMesh` already holds width-keyed
`std::map<int, Vector<RefCountedPtr<LevelData<BaseFab<bool>>>>>` halo and transition masks, defined
at regrid. The structures below are the same shape.

## Passes the filter

**Boundary exposure per cell.** The test is `a_ghostMask.numTargets(cell) > 0`
(`CD_NearestNeighborParticleMergeImplem.H`), a pure function of the three particle ghost masks, which
`Realm::defineParticleGhostMasks()` rebuilds only during regrid. Yet `mergeKDCarve` allocates and
fills an exposure cache **per patch, per merge call, every timestep**
(`CD_KDParticleMergeImplem.H:892`) for data that last changed at the previous regrid. Natural
representation is a width-keyed `LevelData<BaseFab<bool>>` alongside the masks it is derived from.
This is the substantive item; the rest are tidying.

**Per-patch real-space bounds.** `boxRealLo`/`boxRealHi` are recomputed from `dbl[din]` and `dx` at
two sites per merge call. Pure grid geometry, identical every step until the grids move.

**Ghost-neighbour rank sets.** The `getParticleGhostNeighborRanks(width)` prototype from #679 is
regrid-invariant by construction -- built from the masks with zero MPI -- and currently has no owner.
A plain `std::set<int>` per width.

**`dxByLevel`.** Rebuilt in four places (`CD_KDParticleMergeImplem.H:714`, `:1637`;
`CD_NearestNeighborParticleMergeImplem.H:1624`, `:1905`). It is `AmrMesh::getDx()` widened to
`RealVect`. The gain here is that four functions stop re-deriving the same thing, not speed.

## Fails the filter

**Anything templated on the particle payload or backend**: the spatial indices
(`NNSpatialIndex<Packed, Cloud>`, `NNCellSpatialIndex`), the pooled particle maps, the gathered
particle vectors, the kd partition. These are both templated and particle-dependent -- particles move
every step and the merge mutates the set mid-round -- so caching them would be a correctness bug, not
an optimisation.

**Per-call scratch holders**, even though they are not templated. The mergers allocate `EBAMRFAB`
histogram and quota holders per call (`CD_KDParticleMergeImplem.H:726`, `:727`, `:1649`, `:1650`, plus
the nearest-neighbour equivalents). Their *shape* is regrid-invariant, so the allocation could be
hoisted -- but they are mutable working buffers, and putting shared mutable scratch in `AmrMesh` means
two solvers on one realm can quietly collide. These belong to a per-solver owner, not to shared mesh
state. `AmrMesh` should hold invariants, not scratch.

## The ownership problem, and why it points at an operator

Caching the exposure mask inside free functions is unsafe: a cache that outlives a regrid it should
not is a silent wrong-answer bug, and free functions have no invalidation hook. This code already has
that failure mode for ghost masks registered after the grids are built.

`define()`-at-regrid is the invalidation mechanism the rest of the codebase relies on
(`EBGradient`, `EBCoarseFineInterp`, `EBAMRParticleMesh`). Either the invariants live in `Realm`
beside the masks they are derived from, or the mergers become an operator class that owns them. The
scratch holders in the previous section would then belong to that operator rather than to `AmrMesh`,
which resolves the shared-mutable-state objection at the same time.

## Caveat

None of this is measured. The only profile available puts gather at roughly 33% and hash-bound, which
points at neither the exposure recomputation nor the allocations. The exposure item has a clear
argument behind it -- O(cells) of mask queries per timestep for data that changes per regrid -- but
"clear argument" is not "measured win", and the mergers have never been profiled beyond 8 ranks on a
single machine.

Related: #682 (this subsumes its regrid-lifetime caching tier), #679 (gives its prototype an owner),
#683 (same shared-state root cause), #684 (the data-structure cleanup that preceded this).

rmrsk:
# Summary

Closes #685. The particle mergers re-derived, on every call, several quantities that cannot change
between regrids. This moves them to the object whose lifetime already matches -- `Realm`, beside the
particle ghost masks they are derived from -- and removes the duplicated re-derivations.

The branch also carries a signature cleanup across `ParticleManagement`'s merge surface and a
thread-safety audit of the box loops it touches, both described under *Side-effects*, plus merges of
`main` for #689 and #690.

### Background

The substantive item is **boundary exposure per cell**: the test `numTargets(cell) > 0` against the
three particle ghost masks. `mergeKDCarve` allocated and filled a per-patch `BaseFab<signed char>`
exposure cache *every timestep* for data that last changed at the previous regrid, and the
nearest-neighbour gather repeated the same three-mask OR per particle. Alongside it sat smaller
duplications: `dxByLevel` rebuilt in four places, per-patch real-space bounds computed at two sites,
and the kd merges' `EBAMRFAB` scratch holders allocated per call (`AmrMesh::allocate` constructs a
`Copier` per level -- cheap in isolation, non-negligible at scale).

Caching any of this inside the free-function mergers is unsafe: a cache that outlives a regrid it
should not is a silent wrong-answer bug, and free functions have no invalidation hook.
`Realm::defineParticleGhostMasks()` is reached from `regridOperators()` at every regrid and already
owns the masks these quantities are pure functions of, so it is the natural owner -- one definition
point, one invalidation point.

`EBAMRParticleMesh` was considered as the alternative home and rejected: it is owned by `PhaseRealm`
and therefore phase-scoped, while the particle ghost masks and the mergers are realm-scoped and
phase-agnostic. Putting merge state there would key it by an irrelevant phase, duplicate it across
gas/solid, and force it to reach up into `Realm` for data it does not own.

### Solution

`Realm` gains two members, both built inside the existing `defineParticleGhostMasks()` loop over
registered widths and cleared at the top of it alongside the masks:

- `std::map<int, AMRMask> m_particleGhostExposure` -- per level, true in every cell with at least one
  outgoing ghost-shipment target in any direction. Carries **no ghost cells**, deliberately: only
  resident particles are ever tested, and resident means the cell lies in the patch's own box.
- `std::map<int, std::vector<int>> m_particleGhostNeighborRanks` -- for each width, the ranks this rank
  exchanges particle ghosts with, in ascending order, derived with zero MPI. This is the prototype from
  #679, which had no owner; nothing on this branch consumes it yet.

  It is built by flagging ranks in a `std::vector<bool>` and collapsing the flags once, rather than
  inserting into a `std::set`. The masks hold one packed entry per (boundary cell x target) but only a
  handful of distinct ranks, so the set form cost order 10^5-10^6 tree inserts per rank per regrid to
  produce a ~6-element answer. Scanning to collapse the flags yields ascending order for free, which is
  what the eventual point-to-point exchanges need: both sides must walk their shared neighbours in the
  same order for sends and receives to line up. A sorted vector is also what every consumer would
  otherwise copy the set into on each exchange.

`Realm` and `AmrMesh` gain getters for both, modelled on the existing `getParticleGhostMask()`.

Removed as re-derivations of the same invariant data:

- `kdIsBoundaryExposed()` and `isBoundaryExposed()` -- the duplicated one-line mask tests (the former's
  own doc-comment already noted it was a duplicate), plus the per-patch exposure cache.
- The four local `dxByLevel` builds. Two of them disappear outright: every kd use was `dxByLevel[lvl]`
  with `lvl` the enclosing loop variable, so each kd level loop now takes a local `RealVect` and the
  vector is gone from that family entirely. The nearest-neighbour merges genuinely need the whole
  vector -- they index it by each *particle's* level, not the loop's -- so their two entry points build
  it in place. No accessor was added to `AmrMesh` for this: it would have meant a second copy of `m_dx`
  to keep in sync, for something callers can derive.
- The duplicated per-patch real-space bounds, now `detail::kdBoxRealBounds()`.

Finally, the kd merges' per-cell scratch moves to its caller: `mergeKDCarve`/`mergeKDPatch`/
`mergeKDInterior` take the histogram and quota holders as arguments, and `ItoSolver` owns them,
allocating them in `allocate()`/`regrid()` and releasing them in `preRegrid()` exactly like `m_phi` and
the deposition holders.

**New types introduced: none.** Exposure reuses the existing `AMRMask` alias, and the two width-keyed
`std::map`s mirror the three `std::map<int, AMRParticleGhostMask>` members already in `Realm`.

### Side-effects

**Signature cleanup across both merge families.** Adding the scratch parameters interleaved mutable and
non-mutable arguments, which this repository's convention forbids, so the ordering is now
`(<mutables>, <non-mutables>, <optionals>)` throughout `ParticleManagement`'s merge surface -- 22
functions: the three kd entry points, four kd helpers (`buildKDQuotaLeaves`, `kdFillCellHistogram`,
`kdBBox`, `kdBoxRealBounds`), and all sixteen nearest-neighbour functions. Defaulted parameters keep
both their trailing position and their existing relative order, so no call that omits a trailing
optional changes meaning, and no call site gained or lost an argument.

**`AmrMesh` is now passed by const reference.** The mergers call exactly six `AmrMesh` methods --
`getFinestLevel`, `getProbLo`, `getGrids`, `getDx`, `getLevelTiles`, `getParticleGhostExposure` -- and
all six were already `const`. The mutable reference predates this work (even the `a_amr.allocate(...)`
call removed here was itself a `const` member), so this is strictly a tightening.

**Thread-safety audit of the touched box loops.** Six of the seven un-pragma'd `DataIterator` loops in
these files are genuinely unsafe to thread -- shared append buffers, a shared id counter, shared
dedup maps, and (for the neighbour marking) a shared flag array that `std::vector<bool>` packs into
words. Each now records why, instead of leaving the next reader to re-derive it. The one loop that is
safe, summing box sizes for a `reserve()` bound, gains `reduction(+ : combinedCountUpperBound)`.

Other effects:

- The exposure masks cost one `bool` per cell per level per **registered** ghost width. Nothing is
  built unless a width has been registered, so a run that uses no particle ghost masks pays nothing.
- The kd scratch is now held across a timestep rather than allocated and freed within one. It is two
  single-component holders per solver.
- Behaviour is unchanged: particle counts are **identical to `main` across all 24 runs** (six merge
  backends x {2D, 3D} x {serial, 4 ranks}), measured against a baseline built from `main` in a separate
  worktree.
- `main` is merged in twice, for #689 (`ParticleSoA` geometric growth) and #690 (ItoKMC reaction-network
  overhead). Neither overlaps this branch's files.

### Alternative solutions

- **`EBAMRParticleMesh` as the owner** -- rejected for the phase-scoping reason above.
- **A new merge-operator class owning the invariants and the scratch** -- rejected: it would need its
  own `define()`-at-regrid hook duplicating what `Realm` already does, and once the templated state is
  excluded (spatial indices, pooled particle maps, kd partitions -- all particle-dependent and unsafe
  to cache) there are only two non-templated invariants left for it to hold.
- **Leaving the scratch allocated per call** -- rejected because `AmrMesh::allocate()` builds a `Copier`
  per level; cheap at small scale, but it adds up on large runs.
- **Hoisting the scratch above `ItoSolver`** (to `ItoKMCStepper`/`ItoLayout`, one pair instead of one
  per species) -- considered and deferred. It would put two kd-merge buffers into the signature of
  `makeSuperparticles()`, which every merge algorithm shares, including the ones that never touch them.
- **Allocating the scratch lazily**, so a solver that never runs a kd merge pays nothing -- tried and
  dropped. It needed a define/invalidate pair and a state flag to save two single-component holders per
  species; the standard `allocate()`/`preRegrid()` lifecycle is simpler at negligible cost.
- **An `AmrMesh` accessor returning the per-level dx as `RealVect`** -- tried and dropped, see above.
- **Reordering the optional parameters too** (mutable optionals ahead of immutable ones) -- tried and
  reverted: it would have forced call sites that omit a trailing optional to start passing
  `a_crossLevelMergeCount` explicitly, for no benefit.

Verification beyond the 24-run baseline comparison: a temporary DEBUG assertion comparing the cached
exposure bit against the live three-mask OR, for every cell of every patch on every level on every
merge call, ran clean over the whole BrownianWalker merge suite at ghost widths 1 and 2. The neighbour
ranks -- which nothing on this branch reads, so no regression test can exercise them -- were checked by
rebuilding the old `std::set` alongside the new vector and asserting exact equality in order on every
regrid (96 checks over `kd_carve`). Both temporary checks were removed before committing.
`ItoKMC/JSON` runs clean on 4 ranks under `reinitialize`, `kd_carve`, `kd_skin_nn` and `nn_pair_tree`.
The no-MPI build (the CI configuration) and an `OPENMPCC=TRUE` build both compile.

### PR-review checklist

Mandatory:

Optional:

### DIFF
--- a/Docs/Sphinx/source/Source/Particles.rst
+++ b/Docs/Sphinx/source/Source/Particles.rst
@@ -391,7 +391,7 @@ For each source cell a mask lists the destination boxes (grid index and receivin
 There is one mask per direction, and the collection type is
 
 .. literalinclude:: ../../../../Source/AmrMesh/CD_ParticleGhostMask.H
-   :lines: 210
+   :lines: 224
    :language: c++
 
 The masks live on the :ref:`Chap:Realm` and are rebuilt at every regrid.
--- a/Source/AmrMesh/CD_AmrMesh.H
+++ b/Source/AmrMesh/CD_AmrMesh.H
@@ -15,6 +15,9 @@
 #ifndef CD_AMRMESH_H
 #define CD_AMRMESH_H
 
+// Std includes
+#include <set>
+
 // Chombo includes
 #include <DisjointBoxLayout.H>
 #include <ProblemDomain.H>
@@ -1757,6 +1760,16 @@ public:
   getDx() const;
 
   /**
+   * @brief Get spatial resolutions in the per-direction form.
+   * @details The isotropic getDx() widened to RealVect, which is what the point-location and particle
+   * merge code wants. Held rather than rebuilt by each caller: it is a pure function of the domain, so it
+   * does not even change at a regrid.
+   * @return Grid spacings on all levels, as RealVect.
+   */
+  const Vector<RealVect>&
+  getDxAsRealVect() const;
+
+  /**
    * @brief Get refinement ratios
    * @return Refinement ratios between consecutive levels
    */
@@ -1876,6 +1889,30 @@ public:
    */
   const AMRParticleGhostMask&
   getTrivialParticleGhostMask() const;
+
+  /**
+   * @brief Get the particle boundary-exposure mask on a realm for a registered width.
+   * @details True in every cell that has at least one outgoing ghost-shipment target in any direction, i.e.
+   * the OR of the three ghost masks above. Built at regrid, so a consumer can read the bit instead of
+   * re-deriving it per particle per timestep. Carries no ghost cells -- only cells holding a resident
+   * particle may be queried.
+   * @param[in] a_realm Realm name
+   * @param[in] a_width Registered ghost width.
+   * @return Per-level boolean mask over the valid grids.
+   */
+  const AMRMask&
+  getParticleGhostExposure(const std::string& a_realm, const int a_width) const;
+
+  /**
+   * @brief Get the particle ghost neighbor-rank set on a realm for a registered width.
+   * @details Every rank this rank exchanges particle ghosts with, in either direction, at a_width. Lets a
+   * consumer restrict a would-be all-to-all exchange to this set instead.
+   * @param[in] a_realm Realm name
+   * @param[in] a_width Registered ghost width.
+   * @return The neighbor-rank set.
+   */
+  const std::set<int>&
+  getParticleGhostNeighborRanks(const std::string& a_realm, const int a_width) const;
 
   /**
    * @brief Get the EBLevelGrid for a Realm and phase
@@ -2346,6 +2383,13 @@ protected:
    * @brief Level resolutions
    */
   Vector<Real> m_dx;
+
+  /**
+   * @brief Level resolutions in the per-direction form, i.e. m_dx widened to RealVect.
+   * @details Filled beside m_dx in buildDomains() and never touched again -- the grid spacing follows from
+   * the domain, not from the grids, so it survives a regrid unchanged.
+   */
+  Vector<RealVect> m_dxRealVect;
 
   /**
    * @brief Define Realms
--- a/Source/AmrMesh/CD_AmrMesh.H
+++ b/Source/AmrMesh/CD_AmrMesh.H
@@ -1760,16 +1760,6 @@ public:
   getDx() const;
 
   /**
-   * @brief Get spatial resolutions in the per-direction form.
-   * @details The isotropic getDx() widened to RealVect, which is what the point-location and particle
-   * merge code wants. Held rather than rebuilt by each caller: it is a pure function of the domain, so it
-   * does not even change at a regrid.
-   * @return Grid spacings on all levels, as RealVect.
-   */
-  const Vector<RealVect>&
-  getDxAsRealVect() const;
-
-  /**
    * @brief Get refinement ratios
    * @return Refinement ratios between consecutive levels
    */
@@ -2383,13 +2373,6 @@ protected:
    * @brief Level resolutions
    */
   Vector<Real> m_dx;
-
-  /**
-   * @brief Level resolutions in the per-direction form, i.e. m_dx widened to RealVect.
-   * @details Filled beside m_dx in buildDomains() and never touched again -- the grid spacing follows from
-   * the domain, not from the grids, so it survives a regrid unchanged.
-   */
-  Vector<RealVect> m_dxRealVect;
 
   /**
    * @brief Define Realms
--- a/Source/AmrMesh/CD_AmrMesh.H
+++ b/Source/AmrMesh/CD_AmrMesh.H
@@ -15,9 +15,6 @@
 #ifndef CD_AMRMESH_H
 #define CD_AMRMESH_H
 
-// Std includes
-#include <set>
-
 // Chombo includes
 #include <DisjointBoxLayout.H>
 #include <ProblemDomain.H>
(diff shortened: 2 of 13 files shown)


rmrsk:
@claude review

rmrsk:
@claude review

PR #688, opened 2026-08-11
Restrict the bounded particle MPI exchanges to neighbour ranks

rmrsk:


rmrsk:
Will not be merged since costs don't show up in production runs.

PR #690, opened 2026-08-12
Remove per-cell overhead from the ItoKMC reaction-network advance

rmrsk:
# Summary

Removes three sources of avoidable per-cell overhead from `ItoKMCGodunovStepper`'s `advanceReactionNetwork` call: Chombo timers on per-cell and per-reaction functions, redundant cell-sorts and per-cell particle copies, and a copied reaction partition. Static analysis only drove the changes; no algorithm or result changes.

### Background

Profiling `advanceReactionNetwork` statically turned up several costs that are per grid cell, and in places per reaction within a cell.

**Timers on per-cell functions.** All 63 timer sites in `Source/KineticMonteCarlo` sit on functions that are only ever entered per cell — including `KMCSolver::propensities`/`computeDt` and `KMCDualStateReaction::propensity`/`computeCriticalNumberOfReactions`/`getStateChange`, which run once per reaction per cell. With the 8–23 reactions in the shipped chemistries that is on the order of 2000 `CH_TIME` sites entered per grid cell. The same applies to `ItoKMCPhysics::advanceKMC` and the `ItoKMCJSON` per-cell hooks.

These are not compiled out by an optimized build. `CH_TIME` is disabled by `CH_NTIMER`, which the build system emits from exactly one place — `$(subst FALSE,-DCH_NTIMER,$(subst TRUE,,$(USE_TIMER)))` in `Make.rules:277` — so it follows `USE_TIMER` (default `TRUE`, and never overridden in `Lib/Local/` or the Betzy/Fram configs) and is independent of `DEBUG`/`OPT`/`NDEBUG`. A `DEBUG=FALSE OPT=HIGH` build gets `-DNDEBUG` but no `-DCH_NTIMER`.

At runtime, with `CH_TIMER` absent from the environment, `TraceTimer::initializer()` prunes the root timer, so each site returns early. It is *not* free, though: the pruning is a runtime early-out, so every site still costs three out-of-line `TraceTimer` calls (`getTimer`, `start`, `stop`) plus several `omp_get_thread_num()` calls under OpenMP, none of which an optimized build can inline away. That is irrelevant at ordinary call sites — which is why this PR leaves timers alone everywhere else — and is not irrelevant at ~2000 sites per cell.

The residual cost is also asymmetric: the macro only does its work when `omp_get_thread_num() == 0`, so thread 0 carries it alone while the rest of the team waits at the enclosing `omp for` barrier. And with `CH_TIMER=1` — exactly when one is profiling this code — the per-cell sites go fully live and distort the loop being measured, while giving nothing actionable at per-cell granularity.

**Repeated cell-sorting.** The reaction path cell-sorted the same Ito bulk leaves at three separate call sites per step: `ItoKMCGodunovStepper::advance` before the reaction network, `computeReactiveItoParticlesPerCell`, and `reconcileParticles` via `binLeafToCells`. Nothing between them mutates those leaves — the KMC integration works on mesh data (`m_fluidPPC`/`m_fluidYPC`) — so the second and third reproduced a CSR mapping the container already held. Each is a counting sort plus `permute()`, which `aligned_alloc()`s a fresh arena and scatters every column across it (15 columns per `ItoParticle` in 2D, 19 in 3D).

The precondition was in fact already satisfied at every call site in the repo — both callers of `computeReactiveItoParticlesPerCell` sort immediately beforehand. It simply was not expressible, so each helper re-established it defensively.

**Per-cell copies that bought nothing.** `computeReactiveItoParticlesPerCell` and `computeReactiveMeanEnergiesPerCell` went through `binLeafToCells`, copying every particle into per-cell scratch containers, then only accumulated weights and energies out of them.

**A copied partition.** `KMCSolver::partitionReactions` returned its two lists through `std::make_pair` on lvalues, duplicating both allocations and bumping the refcount of every reaction twice more, once per substep per cell.

### Solution

Three commits.

1. **Timers.** Every timer removed from `Source/KineticMonteCarlo` and from the ItoKMC functions that run per cell (`advanceKMC`, `reconcileParticles`, `reconcilePhotons`, `reconcilePhotoionization`, `computeUpstreamPosition`, `removeParticles`, and the `ItoKMCJSON` rate/coefficient hooks). Stepper-level timers are untouched, so the `compute_ppc` / `integrate_network` / `copies` / `reconcile_particles` / `reconcile_cdr` split reports as before. `CH_Timer.H` includes are replaced with the headers actually used. Also moves rather than copies in `partitionReactions`.

2. **Read in place.** The two per-cell-count kernels index the leaf's CSR ranges via `cellRange(c)` instead of extracting copies. `binLeafToCells` stays for `reconcileParticles` and the secondary-emission path, which need mutable per-cell containers to hand to the physics interface.

3. **Idempotent sort.** `ParticleSoA` records the `(box, dx, probLo)` its CSR mapping was built against and `sortByCell` returns early when asked for the same domain, exposed as `isSortedAgainst()`. This collapses the chain to **one** real sort per step for the whole-patch mergers (`kd_patch`, `kd_carve`, `kd_skin_nn`, `nn_pair_*`), and to two for the per-cell mergers, which sort again inside `applyCellMerger` after `reconcileParticles` has rebuilt the leaves.

An early-out is only as sound as the invalidation of every mutating path, and two paths moved particles without clearing `m_sorted`: `setPosition`, and the non-const `positionColumn()`, whose raw `double*` the transport kernels write through. That was harmless while `sortByCell` was unconditional but would have become a silent wrong-answer bug here, so the flag is now cleared when the writable column pointer is handed out — the write itself is not observable to the container, the handout is. Read-only kernels keep the sort by taking the existing const overload.

To keep the invariant checkable rather than merely asserted, `sortByCell` verifies the mapping against the current positions on every early-out via `checkCellSort()`, which recomputes each particle's cell key and confirms it lies in that cell's range. It shares `cellKey()` with the sort itself so the two cannot disagree, is O(N), and compiles out with the assertion in optimized builds.

### Side-effects

- **No new types.** The change adds three data members to `ParticleSoA` (`m_sortBox`, `m_sortDx`, `m_sortProbLo`) and two member functions (`isSortedAgainst`, public, 1 use in `sortByCell` plus external callers; `checkCellSort`, protected, 1 use). No new class, struct, or container.
- **`positionColumn()`'s non-const overload now has a side effect** — it clears the sort flag. Callers that only read through it (e.g. `ItoKMCGodunovStepperImplem.H:829`, which copies position into `old_x`) now drop the sort unnecessarily. That site runs during transport, before the sort, so it costs nothing today, but the pattern is worth a later sweep toward the const overload.
- `checkCellSort` makes DEBUG builds do O(N) verification work per skipped sort. Compiled out in optimized builds.
- `Particles.rst`'s `extractCell` literalinclude range updated for the shifted line numbers; `sphinx -W --keep-going` is clean.

### Alternative solutions

- **A bare `if (isSorted()) return;` guard** would have been the same speedup for far less code, but `m_sorted` records only *that* some mapping is valid, not *which* domain it was built against. That makes the invariant less checkable than it is today, and would silently misbehave the first time a leaf is sorted against a different box.
- **Making `rebuildLeafFromCells` mark the rebuilt leaf sorted** would remove the last sort for the per-cell mergers. `reconcileParticles` concatenates its per-cell scratch in `Box::index` order, so the output is already physically CSR-ordered, and the offsets are a running sum in a loop that already runs — nearly free. Rejected for now: it buys **zero** for `kd_patch` and the other whole-patch mergers (which never sort the bulk leaf — only the per-cell family routes through `applyCellMerger`), applies only on `merge_interval` steps, and promotes three undocumented facts to load-bearing invariants (photoionization products inherit the absorbed photon's position; those photons were binned to the same cell; `cellKey` clamps identically in the sort and the verifier). `ItoKMCStepper::reconcilePhotoionization()` appends products to the leaf tail and would violate it outright — it is currently dead code, but it is a loaded gun next to the invariant.
- **Replacing `ReactionList`'s `std::shared_ptr` with a non-owning pointer** would remove the remaining 2R atomic refcount operations per substep per cell. Rejected: the refcounts are per-thread and uncontended, so this is roughly 1–5% of what the per-cell KMC work costs, and it would break the `updateReactionRates` signature — a public extension point that `ItoKMCJSON` overrides.

### Not addressed here

The dominant per-cell cost found during this analysis is untouched and is 1–2 orders of magnitude larger than anything in this PR: the time-step-limiting tail of `ItoKMCPhysics::advanceKMC` is O(R²) heap allocations and O(R²·U) `std::map` lookups per cell, and runs unconditionally in cells with zero particles. Specifically — `auto S = m_kmcState` is copied above the `N < max` test that is its only consumer; `KMCSolver::computeDt` rebuilds a run-invariant reactant set (concatenating by-value `std::list`s, then sort + unique) on every call and looks up a fixed stoichiometry matrix through `std::map::find` + `at`; and `KMCDualStateReaction::propensity` copies the whole state vector because `auto` drops the reference from `const State&`. Worth a follow-up PR.

### Testing

- Compile-checked in 2D and 3D, DEBUG and OPT, across `ItoKMCJSON.cpp`, `ItoKMCPhysics.cpp`, both standalone KMC test drivers, and `Exec/Tests/ItoKMC/JSON`.
- `Exec/Tests/ItoKMC/JSON` regression run in a DEBUG build (assertions live, so `checkCellSort` verifies every early-out): 26 steps with regrids every 5, particle count 189 decaying to 1, clean exit, no assertion failures and no NaNs.
- `pre-commit run` clean on all changed files (clang-format, reuse, codespell, cppcheck, doxygen).
- `python3 -m sphinx -W --keep-going` clean.

### PR-review checklist

Mandatory:

Optional:

### DIFF
--- a/Docs/Sphinx/source/Source/Particles.rst
+++ b/Docs/Sphinx/source/Source/Particles.rst
@@ -726,7 +726,7 @@ The recommended pattern for the per-cell family is to cell-sort the leaf, extrac
 ``ParticleSoA<P>::extractCell`` performs the per-cell extraction
 
 .. literalinclude:: ../../../../Source/Particle/CD_ParticleSoA.H
-   :lines: 1506-1528
+   :lines: 1541-1563
    :dedent: 2
    :language: c++
 
--- a/Physics/ItoKMC/CD_ItoKMCPhysicsImplem.H
+++ b/Physics/ItoKMC/CD_ItoKMCPhysicsImplem.H
@@ -348,8 +348,6 @@ ItoKMCPhysics::advanceKMC(Vector<FPR>&            a_numParticles,
                           const Real              a_dx,
                           const Real              a_kappa) const
 {
-  CH_TIME("ItoKMCPhysics::advanceKMC");
-
   // Note: This is called PER GRID CELL, i.e. within OpenMP parallel regions. For this reason the KMC solver
   //       must be defined through defineKMC() (which must be later killed).
   CH_assert(m_isDefined);
@@ -523,10 +521,6 @@ ItoKMCPhysics::reconcileParticles(Vector<ParticleSoA<ItoParticle>*>& a_particles
                                   const Real                         a_dx,
                                   const Real                         a_kappa) const noexcept
 {
-  CH_TIMERS("ItoKMCPhysics::reconcileParticles");
-  CH_TIMER("ItoKMCPhysics::reconcileParticles::compute_downstream", t1);
-  CH_TIMER("ItoKMCPhysics::reconcileParticles::particle_placement", t2);
-
   CH_assert(m_isDefined);
   CH_assert(a_particles.size() == a_newNumParticles.size());
   CH_assert(a_oldNumParticles.size() == a_newNumParticles.size());
@@ -580,7 +574,6 @@ ItoKMCPhysics::reconcileParticles(Vector<ParticleSoA<ItoParticle>*>& a_particles
                                                   a_dx);
   }
 
-  CH_START(t2);
   for (int i = 0; i < a_particles.size(); i++) {
     const long long diff = static_cast<long long>(a_newNumParticles[i] - a_oldNumParticles[i]);
 
@@ -653,7 +646,6 @@ ItoKMCPhysics::reconcileParticles(Vector<ParticleSoA<ItoParticle>*>& a_particles
       this->removeParticles(*a_particles[i], -diff);
     }
   }
-  CH_STOP(t2);
 }
 
 inline bool
@@ -666,8 +658,6 @@ ItoKMCPhysics::computeUpstreamPosition(RealVect&                       a_pos,
                                        const RealVect&                 a_cellPos,
                                        const Real&                     a_dx) const noexcept
 {
-  CH_TIME("ItoKMCPhysics::computeUpstreamPosition");
-
   CH_assert(a_dx > 0.0);
 
   a_pos = RealVect::Zero;
@@ -712,8 +702,6 @@ ItoKMCPhysics::computeUpstreamPosition(RealVect&                       a_pos,
 inline void
 ItoKMCPhysics::removeParticles(ParticleSoA<ItoParticle>& a_particles, const long long a_numParticlesToRemove) const
 {
-  CH_TIME("ItoKMCPhysics::removeParticles");
-
   constexpr long long zero = 0LL;
 
   CH_assert(m_isDefined);
@@ -786,8 +774,6 @@ ItoKMCPhysics::reconcilePhotons(Vector<ParticleSoA<Photon>*>& a_newPhotons,
                                 const Real                    a_dx,
                                 const Real                    a_kappa) const noexcept
 {
-  CH_TIME("ItoKMCPhysics::reconcilePhotons");
-
   CH_assert(m_isDefined);
 
   for (int i = 0; i < a_newPhotons.size(); i++) {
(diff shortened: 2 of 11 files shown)


rmrsk:
@claude review

PR #691, opened 2026-08-12
Remove the O(R^2) per-cell cost from the ItoKMC time step calculation

rmrsk:
# Summary

Removes the O(R²) per-cell cost from the ItoKMC time step calculation, where R is the number of reactions. This is the follow-up #690 named in its "Not addressed here" section. Per-cell cost drops 2.5×–5.8× depending on chemistry size, and the scaling in R changes from ~R^1.8 to ~R^1.25. Results are unchanged.

Also skips whole patches with no cut cells in `fillSecondaryEmissionEB`.

### Background

#690 removed per-cell timer and sort overhead from `advanceReactionNetwork` and recorded what it left behind:

> The dominant per-cell cost found during this analysis is untouched and is 1–2 orders of magnitude larger than anything in this PR: the time-step-limiting tail of `ItoKMCPhysics::advanceKMC` is O(R²) heap allocations and O(R²·U) `std::map` lookups per cell, and runs unconditionally in cells with zero particles.

Profiling confirms it. A driver built on the `Exec/Tests/KineticMonteCarlo/Avalanche` pattern — using the types `ItoKMCPhysics` actually instantiates (`KMCDualState`/`KMCDualStateReaction` with `FPR = Real`), driving the real `KMCSolver::advanceHybrid` with `hybrid_midpoint` and the solver parameters from `regression2d.inputs` — attributes the per-cell cost as follows, for the 20-reaction air chemistry in `Exec/Tests/CdrPlasma/JSON` with its wildcards expanded:

| | µs/cell | share |
|---|---|---|
| `KMCSolver::computeDt` | 17.3 | **74%** |
| `propensities` | 3.4 | 15% |
| `computeCriticalNumberOfReactions` | 0.45 | 2% |
| state copies in the tail | 0.39 | 2% |
| the KMC advance itself | 1.1 | 5% |

`computeDt` is entered 16 times per cell for a 20-reaction set: once for the cell state, then once per reaction that can fire.

Three structures are rebuilt on every one of those calls even though none of them depends on the state:

- **The distinct reactant set.** `computeDt` and `getNonCriticalTimeStep` assembled it by concatenating each reaction's `getReactants()` — which returned a `std::list<size_t>` **by value**, so one heap node per reactant per reaction per call — then `sort()`ing and `unique()`ing the resulting linked list.
- **The stoichiometry.** The innermost loop called `getStateChange(reactant)`, which did `std::map::find` followed by `std::map::at`: two red-black-tree walks per lookup, |U|·R times per call.
- **The propensity's working state.** `KMCDualStateReaction::propensity` did `auto reactiveState = a_state.getReactiveState()` — `auto` drops the reference from the `const State&` — copying the whole state vector on every call, purely so it could decrement its own copy while forming the falling factorial.

The `advanceKMC` tail additionally copied `m_kmcState` above the `N < max` test that is its only consumer, so the copy was paid for every reaction including those that never enter the branch, and returned its propensities by value.

`getNonCriticalTimeStep` carries the same reactant-set and stoichiometry pattern and is called once per reactive substep from inside `advanceHybrid`, so this is not confined to the time step tail.

### Solution

- `KMCSolver` gains four `mutable` scratch buffers and `gatherDistinctReactants()`. The distinct reactant set is collected in a single pass using membership stamps — no sort, no allocation — and `computeDt`/`getNonCriticalTimeStep` accumulate reaction-major into per-species scratch before reducing. The sums are the same as reactant-major; only the loop order differs.
- `KMCDualStateReaction` gains `m_reactiveStateChangeDense` (a dense mirror of `m_reactiveStateChange`, built once in `computeStateChanges()`) so `getStateChange` is an indexed read, and `m_lhsOrders` (the reactant tally that already fixes `m_propensityFactor`, compacted to `(species, order)` pairs) so `propensity` forms the falling factorial straight off the incoming state.
- `getReactants()` returns by const reference in both `KMCDualStateReaction` and `KMCSingleStateReaction`.
- The `advanceKMC` tail reuses a thread-local propensity buffer and a thread-local perturbed state, and copies the state only inside the branch that reads it.
- `fillSecondaryEmissionEB` fetches its `VoFIterator` first and skips the patch when it is empty.

Measured against the parent commit, same machine, `OPT=HIGH`:

| chemistry | reactions | before | after | speedup |
|---|---|---|---|---|
| `simple_air` (`Exec/Tests/ItoKMC/JSON`) | 4 | 1.174 µs/cell | 0.468 | **2.5×** |
| AirBasic (`Exec/Examples/ItoKMC/AirBasic`) | 5 | 1.719 µs/cell | 0.614 | **2.8×** |
| air (`Exec/Tests/CdrPlasma/JSON`, `@` expanded) | 20 | 29.191 µs/cell | 5.026 | **5.8×** |

Cost against reaction count, 8 reactive species:

| reactions | before µs/cell | after |
|---|---|---|
| 4 | 1.43 | 0.44 |
| 8 | 4.55 | 0.90 |
| 20 | 24.21 | 2.78 |
| 40 | 87.85 | 7.86 |

~R^1.8 before, ~R^1.25 after.

### Side-effects

- **`KMCSolver`'s `const` member functions are no longer reentrant.** The scratch is `mutable`, so a single `KMCSolver` must not be shared across threads. `ItoKMCPhysics` satisfies this by holding its solver as `static thread_local` (see `defineKMC`), and the constraint is documented on each member. This is the one deliberate trade in the PR: it is what makes the hot path allocation-free.
- **`getReactants()`'s return type changed** from `std::list<size_t>` to `const std::list<size_t>&` in both reaction classes. Source-compatible for every use in the repo (`auto x = r->getReactants()` still copies); it would affect out-of-tree code only if something bound the result in a way that relied on it being a temporary.
- **`m_reactiveStateChange` is now mirrored, not replaced.** The map remains the authoritative store and is still iterated by `advanceState` and `computeCriticalNumberOfReactions`, which are sparse walks and measured at ~2% each. Keeping both means the dense table must be rebuilt whenever the map changes; `computeStateChanges()` is the only place that writes it.
- The `fillSecondaryEmissionEB` skip changes nothing inside a processed patch — the body is unchanged, so the consume-once/overwrite behaviour in multi-valued cells (where several VoFs share one `IntVect`, and the per-cell scratch is keyed by `IntVect`) is preserved exactly.
- No new types, classes, or containers. The change adds two data members to `KMCDualStateReaction`, four scratch members plus one protected helper to `KMCSolver`, and two `static thread_local` members to `ItoKMCPhysics`.

### Alternative solutions

- **Caching the reactant set and stoichiometry on the solver at `define()` time** would remove even the single gathering pass. Rejected because both routines are handed a *sub*-list: `getNonCriticalTimeStep` receives the non-critical partition, which changes per cell and per substep, so a table keyed on the full reaction list would not serve it. Hanging the dense stoichiometry off each reaction works for any sub-list.
- **A scratch workspace passed in by the caller** would keep `KMCSolver` reentrant with no allocation. Rejected because it changes the signatures of `computeDt` and `getNonCriticalTimeStep`, which are public and called directly by the standalone drivers and by user code.
- **Local `std::vector` scratch** would also keep the class reentrant, at two small allocations per call (~32 per cell at R=20). Rejected as the worst of both: it keeps allocation in the hot path while still being slower than the mutable-member form, for a reentrancy guarantee nothing in the repo uses.
- **An early-out for cells whose reactive state is entirely zero.** Rejected: it is only sound while every reaction has at least one reactant. A zero-order reaction such as `<nothing> -> e + O2+` (background ionisation) has a non-zero propensity in an empty cell, so the guard would be a latent trap the first time such a reaction is added. After the hoists an empty cell costs little enough that it is not worth the risk.
- **Rewriting `fillSecondaryEmissionEB` to size its scratch by cut cells rather than by `box.numPts()`.** Deferred: it changes behaviour in multi-valued cells (`extractCell` copies rather than consumes, so every VoF sees the cell's particles, and appending accumulates rather than overwrites). Both the old and new semantics are arbitrary — the cell-sort keys on `IntVect` and cannot say which VoF a particle belongs to — so this belongs in its own change with that question settled deliberately. The patch skip captures most of the win because EB-touching patches are a small minority.

### Testing

Both suites were run against a build of the parent commit in a separate tree, with identical inputs and `Random.seed` pinned, and compared line for line after filtering wall-clock timings, OS memory readouts, and the working-directory banner.

- `Exec/Tests/ItoKMC/JSON`, 4 MPI ranks, 16 time steps: **539 physics lines identical on all 4 ranks**.
- `Exec/Examples/ItoKMC/PartialDischarge`, 4 MPI ranks, 25 time steps: **1199 physics lines identical on all 4 ranks**. This configuration is the one whose chemistry has a `dielectric emission` block; `Qsurf` grows monotonically across the run (24 distinct values), confirming the secondary-emission path is exercised rather than merely compiled.
- The profiling driver additionally cross-checks `dt` against an independent implementation over 61,440 cell evaluations spanning three chemistries and both empty and populated cells: **0 differ, maximum relative difference 0**.
- Compile-checked in 2-D and 3-D: `CD_ItoKMCPhysics.cpp`, `CD_ItoKMCJSON.cpp`, `Exec/Tests/ItoKMC/JSON/main.cpp`, and both standalone KMC drivers (`Avalanche`, `Schlogl`), which exercise the `KMCSingleStateReaction` path.
- `pre-commit run` clean on all changed files (clang-format, reuse, codespell, cppcheck, doxygen).

Not done: no Sphinx documentation covers these internals, so none was updated; no `.rst` `literalinclude` references the changed line ranges.

### PR-review checklist

Mandatory:

Optional:

### DIFF
--- a/Physics/ItoKMC/CD_ItoKMCPhysics.H
+++ b/Physics/ItoKMC/CD_ItoKMCPhysics.H
@@ -479,6 +479,21 @@ protected:
   static thread_local KMCState m_kmcState;
 
   /**
+   * @brief Perturbed KMC state used by the time step tail of advanceKMC.
+   * @details The tail probes each reaction by firing it to its critical number and re-evaluating
+   * the time step on the result. It needs a state to perturb without disturbing m_kmcState, and
+   * reuses this one so that the probe does not allocate once per reaction per grid cell.
+   */
+  static thread_local KMCState m_kmcStateScratch;
+
+  /**
+   * @brief Propensity buffer used by the time step tail of advanceKMC.
+   * @details Reused across the tail's probes so the propensities are not returned by value, which
+   * allocated once per probe, i.e. once per reaction per grid cell.
+   */
+  static thread_local std::vector<Real> m_kmcPropensityScratch;
+
+  /**
    * @brief Thread-local copies of KMC reactions used in advanceReactionNetwork.
    * @details Set up via defineKMC() to ensure OpenMP thread safety; depopulated by killKMC().
    */
--- a/Physics/ItoKMC/CD_ItoKMCPhysics.cpp
+++ b/Physics/ItoKMC/CD_ItoKMCPhysics.cpp
@@ -23,6 +23,8 @@ thread_local bool                                            ItoKMCPhysics::m_ha
 thread_local KMCSolverType                                   ItoKMCPhysics::m_kmcSolver;
 thread_local KMCState                                        ItoKMCPhysics::m_kmcState;
 thread_local std::vector<std::shared_ptr<const KMCReaction>> ItoKMCPhysics::m_kmcReactionsThreadLocal;
+thread_local KMCState                                        ItoKMCPhysics::m_kmcStateScratch;
+thread_local std::vector<Real>                               ItoKMCPhysics::m_kmcPropensityScratch;
 
 Vector<std::string>
 ItoKMCPhysics::getPlotVariableNames() const noexcept
--- a/Physics/ItoKMC/CD_ItoKMCPhysicsImplem.H
+++ b/Physics/ItoKMC/CD_ItoKMCPhysicsImplem.H
@@ -122,6 +122,9 @@ ItoKMCPhysics::defineKMC() const noexcept
   m_kmcSolver.define(m_kmcReactionsThreadLocal);
   m_kmcSolver.setSolverParameters(m_Ncrit, m_NSSA, m_maxIter, m_eps, m_SSAlim, m_exitTol);
   m_kmcState.define(m_itoSpecies.size() + m_cdrSpecies.size(), m_rtSpecies.size());
+  m_kmcStateScratch.define(m_itoSpecies.size() + m_cdrSpecies.size(), m_rtSpecies.size());
+
+  m_kmcPropensityScratch.resize(m_kmcReactionsThreadLocal.size());
 
   m_hasKMCSolver = true;
 }
@@ -136,6 +139,9 @@ ItoKMCPhysics::killKMC() const noexcept
   m_kmcReactionsThreadLocal.resize(0);
   m_kmcSolver.define(m_kmcReactionsThreadLocal);
   m_kmcState.define(0, 0);
+  m_kmcStateScratch.define(0, 0);
+
+  m_kmcPropensityScratch.resize(0);
 
   m_hasKMCSolver = false;
 }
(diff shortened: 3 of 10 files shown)


rmrsk:
@claude review

PR #692, opened 2026-08-13
Remove the particle ghost neighbour-rank facility

rmrsk:
# Summary

Removes the particle ghost neighbour-rank facility. It was added in #685 as an invariant for #679's
point-to-point exchanges to consume; #679 is not being merged, so nothing consumes it.

### Background

`Realm` computed, for each registered particle ghost width, the set of ranks this rank exchanges
particle ghosts with -- the union over every box it owns, at every level, of the target ranks appearing
in that width's same-level / fine-to-coarse / coarse-to-fine masks. It is derivable with no MPI at all,
because ghost adjacency is symmetric, and that is what made it attractive: a rank could discover its
neighbours for free and then talk only to them.

It landed in #685 because it was one of that issue's regrid invariants and had no owner. The intended
consumer was #679, which replaces the full-communicator `MPI_Alltoall`/`Alltoallv` in ghost-fill and in
the mergers' claim/verdict/commit rounds with `Isend`/`Irecv` restricted to that set.

That consumer is not arriving. The all-to-all does not show up in production profiles at 1k ranks, so
the exchange rewrite has no measured benefit to justify its complexity, and #679 is being closed
unmerged. Two supporting measurements, for the record:

- The neighbour count does saturate as the communicator grows -- 3.0 of 3 possible at 4 ranks, 5.7 of 7
  at 8, 5.9 of 15 at 16 -- so the bounded-neighbour premise itself holds.
- But a desktop A/B of the two exchange implementations was inconclusive: run-to-run variance on a
  single build spanned 0.105-0.229 s for the merge exchange timer at 16 ranks, against a branch
  difference of 3-8%.

### Solution

Delete the facility and the work that built it:

- `Realm::m_particleGhostNeighborRanks` and `Realm::getParticleGhostNeighborRanks`
- `AmrMesh::getParticleGhostNeighborRanks`
- `ParticleGhostMask::markTargetRanks`
- in `Realm::defineParticleGhostMasks()`: the per-rank flag array, the per-box marking loop, and the
  collapse pass that turned the flags into a sorted vector

This removes a per-regrid cost nobody was paying for. The marking loop visited every packed target of
all three masks -- one entry per (boundary cell x target) -- on every box of every level, to produce a
result with no reader.

**The boundary-exposure mask that #685 added alongside it stays.** That one has consumers in both merge
families (`mergeKDCarve` and `gatherMergeParticles` both read it instead of re-deriving the three-mask
OR per particle), and it is the item #685 called substantive.

### Side-effects

None intended, and none observed: this is a pure deletion of code with no callers. `kd_carve` and
`nn_pair_tree` reproduce their particle counts exactly on four ranks.

One documentation fix comes along with it. `Particles.rst` pulls the `AMRParticleGhostMask` alias out of
`CD_ParticleGhostMask.H` by absolute line number; deleting `markTargetRanks` shifted the alias, so the
directive is repointed from line 224 to 210. Every line-addressed `literalinclude` into a file this PR
touches (13 of them) was checked by rendering each directive against its own revision and comparing at
the word level; only that one moved, and it now renders what it rendered before.

### Alternative solutions

- **Keep the facility on the chance that a future consumer wants it.** Rejected: it is not free -- it
  costs a full sweep of every mask's packed targets at every regrid -- and the argument for its consumer
  is exactly what failed to hold up. If the point-to-point work is ever revisited, this is ~60 lines to
  reinstate from the history, against a cost paid every regrid until then.
- **Keep it but stop building it unless asked** (build on first request, cache per regrid). Rejected as
  machinery in service of a hypothetical caller.

### PR-review checklist

Mandatory:

Optional:

### DIFF
--- a/Docs/Sphinx/source/Source/Particles.rst
+++ b/Docs/Sphinx/source/Source/Particles.rst
@@ -391,7 +391,7 @@ For each source cell a mask lists the destination boxes (grid index and receivin
 There is one mask per direction, and the collection type is
 
 .. literalinclude:: ../../../../Source/AmrMesh/CD_ParticleGhostMask.H
-   :lines: 224
+   :lines: 210
    :language: c++
 
 The masks live on the :ref:`Chap:Realm` and are rebuilt at every regrid.
--- a/Source/AmrMesh/CD_AmrMesh.H
+++ b/Source/AmrMesh/CD_AmrMesh.H
@@ -1891,17 +1891,6 @@ public:
   getParticleGhostExposure(const std::string& a_realm, const int a_width) const;
 
   /**
-   * @brief Get the particle ghost neighbor ranks on a realm for a registered width.
-   * @details Every rank this rank exchanges particle ghosts with, in either direction, at a_width. Lets a
-   * consumer restrict a would-be all-to-all exchange to this set instead.
-   * @param[in] a_realm Realm name
-   * @param[in] a_width Registered ghost width.
-   * @return The neighbor ranks, in ascending order.
-   */
-  const std::vector<int>&
-  getParticleGhostNeighborRanks(const std::string& a_realm, const int a_width) const;
-
-  /**
    * @brief Get the EBLevelGrid for a Realm and phase
    * @param[in] a_realm Realm name
    * @param[in] a_phase Phase (gas or solid)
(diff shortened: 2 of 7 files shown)


rmrsk:
@claude review

Issue #697, opened 2026-08-14
Consolidate the container-owned particle state: the identity triple is 13 bytes of unreachable width


rmrsk:
## Summary

Every `ParticleSoA` carries five container-owned columns regardless of what the particle is: position,
weight, `particleID`, `rankID` and `ghostType`. In 3-D that is 45 bytes, of which 13 are the
identity triple, and the widths are far larger than anything reachable.

This matters more than it looks, for two reasons. Container columns are paid by every particle type, so
a saving here applies everywhere at once. And for `NoPayload` particles they are the *entire*
footprint — there is no payload to trim — which makes this the only lever that reaches them.

## Where the bytes are

Container-owned columns, 3-D:

| column | type | bytes | share |
|---|---|---|---|
| position | 3 x double | 24 | 53% |
| weight | double | 8 | 18% |
| particleID | int64 | 8 | 18% |
| rankID | int32 | 4 | 9% |
| ghostType | uint8 | 1 | 2% |

The identity triple is 13 B of the 45. Realistic bounds are nowhere near their widths:

| field | current | reachable | needed |
|---|---|---|---|
| rankID | 32 bits | 2^21 = 2.1 M ranks | ~21 bits |
| ghostType | 8 bits | a handful of enumerators | ~3 bits |
| particleID | 64 bits | 2^40 = 1.1e12 ids | ~40 bits |

Those fit in a single 64-bit column: **13 B -> 8 B**.

`particleID` cannot simply be dropped. It is load-bearing for the merge algorithms, which carry it as
`globalID` and use it to look up consumed particles (`CD_KDParticleMergeImplem.H`,
`CD_NearestNeighborParticleMergeImplem.H`). It can be narrowed, not removed.

Position is the floor. 24 B of the 45 is coordinates, and single precision over a 0.1 m domain with
~10 um cells leaves too little margin to be worth the risk. Below about 36 B per point particle there
is nothing left to take.

## Why the NoPayload containers make this worth doing

`ItoKMCGodunovStepper` holds three `Vector<RefCountedPtr<ParticleContainer<NoPayload>>>` —
`m_rhoDaggerParticles`, `m_conductivityParticles` and `m_irregularParticles` — filled roughly one
point particle per Ito particle. At 45 B each that is **135 B per Ito particle, about 43% of the
particle arena**, and every byte of it is container-owned.

On a 1024-rank production run the particle arenas peaked at 789 MB on the worst rank. Packing the
[...]

Issue #696, opened 2026-08-14
Regrid de-refinement transiently holds 2^D times the target particles per cell


rmrsk:
## Summary

At a regrid the Ito particles are redistributed onto the new grids in full `ItoParticle` form, and the
super-particle merge runs only afterwards. When a level is de-refined the particles from `2^D` fine
cells land in a single coarse cell, so between the redistribution and the merge the containers hold up
to `2^D` times the target particles per cell — eight times in 3-D — in real, resident arenas.

This is a peak rather than a leak. It is bounded, it is merged away, and it never appears in any
cumulative series. But peaks are what get a job killed, and this one is the largest single term in the
particle footprint.

## Evidence

From a 1024-rank ItoKMC run on Betzy, instrumented with per-kind particle memory counters:

| | |
|---|---|
| peak particle arenas, worst rank | **789 MB** |
| peak particle buffers (payload in flight), worst rank | 293 MB |
| peak resident set, worst rank | **1449 MB** |
| steady-state arenas on that rank | ~260 MB |

The jump occurs in a single step, at a regrid that coarsened the grid from 1315 to 1131 boxes. The
global particle count *falls* by 10 M across that step — that is the merge afterwards removing what the
de-refinement piled up. The observed factor is about 3x rather than the full 8x because only part of
the domain de-refines.

Two things rule out the alternatives. The peak is 94% arenas rather than exchange buffers, so it is
resident particle data and not payload in flight. And measured against live particle counts the arenas
carry only 1.1–2.1x what the particles in them need, so it is not retained capacity or allocator slack
either — the population itself is transiently larger.

## Why the ordering causes it

`ItoKMCStepper::regrid` redistributes first and merges second:

```cpp
m_ito->regrid(a_lmin, a_oldFinestLevel, a_newFinestLevel);   // particles land on the new grids
...
if (!mergeIsOff) {
  m_ito->makeSuperparticles(ItoSolver::WhichContainer::Bulk, m_particlesPerCell);
}
```

`ParticleContainer::regrid` fills every patch in one call, so nothing can merge until all of them are
full. Note also that `regrid_superparticles = none` leaves the pile-up standing until the next step's
merge, which is a memory cost worth documenting on that option.

## Proposal

Carry a cheaper particle across the regrid, and merge in the solver that owns the physics.

- `ItoSolver::preRegrid` copies into a reduced holder that keeps only genuine state: position, weight,
[...]

PR #695, opened 2026-08-14
Update Chombo

rmrsk:
# Summary

This PR updates Chombo, as we found a memory tracking bug that caused Vector to pin its counters too high, inflating the reported (but not real) memory usage.

### Background

In a Betzy run, the memory usage dropped by 10x (seemingly) upon a restart. An investigation tracked this down to a bad counter in Chombo.

### Solution

Decrement Vector in the move constructor -- ensuring proper counting.

### Side-effects

No actual change to how Vector works; only how its tracked by the memory counter. No side effects expected.

### Alternative solutions 

None considered.

### PR-review checklist

Mandatory:

- [x] I have run the test suite and made sure that it passed.
- [x] I have run CheckDocs.py and fixed potentially broken literalincludes.
- [x] I have added all relevant user documentation to Sphinx.
- [x] I have added all relevant APIs to the doxygen documentation.
- [x] I have added appropriate labels to this PR.
- [x] I have added/revised proper licensing and copyright information.
- [x] I have run a PR review using `@claude review`.

Optional:

- [x] I have run valgrind to make sure that this PR does not cause memory leaks. 
- [ ] I have run clang-tidy and fixed all reported errors.
- [ ] I have made sure that this PR does not change benchmark files (unless it is intended to do so).

Issue #693, opened 2026-08-13
Vector<T>::operator=(Vector<T>&&) leaves the destination's element count on the memory-tracking books


rmrsk:
## Summary

Chombo's `Vector<T>::operator=(Vector<T>&&)` performs no memory-tracking accounting, so every
move-assignment onto a non-empty `Vector` permanently inflates the reported "unfreed memory" by the
destination's prior size. **No memory is actually retained** — this is a counter bug — but the error
is one-directional, so the reported figure accumulates with the number of move-assignments the
process has performed, i.e. with run history.

This is what produced the long-standing "ItoKMC leaks memory without bound at a fixed grid" symptom,
and it is very likely most of the "a restart from a checkpoint reports 10.4x less unfreed memory for
the same physical state" observation that anchored that investigation — a restart has performed
almost no move-assignments yet.

The code lives in the `Chombo-3.3` submodule; filing here because all the evidence and all the
affected runs are chombo-discharge.

## The bug

`Chombo-3.3/lib/src/BaseTools/Vector.H:127`

```cpp
inline
Vector<T>& operator=(Vector<T>&& a_invec)
{
  v=std::move(a_invec.v);          // no increment(), no decrement()
  return *this;
}
```

Compare the copy-assign directly beneath it, which brackets the assignment with
`decrement(size())` / `increment(size())`.

Walk the counter through `dst = std::move(src)`, `dst` holding N elements, `src` holding M:

| | real memory | tracked counter |
|---|---|---|
| before | dst: N, src: M | N + M |
| after the move | dst: M, src: 0 | N + M (unchanged) |
| `src` destroyed | — | −0 (it is empty; `~Vector` decrements `size()`) |
| `dst` destroyed | M freed | −M |
| **net left on the books** | **0** | **+N, forever** |

### Why it was written this way

The move *constructor* immediately above deliberately skips accounting:

```cpp
Vector(Vector<T>&& a_invec) noexcept : v(std::move(a_invec.v))
{
  //do not alter counting, this is a move operation
}
```

and for **construction that is correct** — the destination is brand new, so it holds nothing to
release. The same reasoning was evidently carried across to move-*assignment*, where it is wrong,
because there the destination is not empty.

I audited the siblings: the move constructor and `swap` are both correct by construction (`swap`
leaves the total unchanged and each destructor decrements its own current size). **Move-assignment is
the only gap.**

## Evidence

Fix is two lines — `decrement(size())` before the move. Saved as
[...]

PR #698, opened 2026-08-14
Fix the multigrid preconditioner leak, slim the particle exchange, and rework the memory report

### LINKED ISSUE #694: Particle remap holds three copies of every outbound particle, and its exchange silently caps at 2 GB in OPT builds
## Summary

`ParticleContainer::distributeFromPool` holds **three simultaneous copies** of every outbound
particle during the cross-rank exchange, and the exchange itself is silently capped at 2 GB per rank
in optimised builds. Both live in `Source/Particle/CD_ParticleContainerImplem.H`.

At the scale of the 512-core ItoKMC run that died with SIGKILL, the copies are measured in gigabytes
per rank, so this is a plausible contributor to that failure in a way that the tracked-memory
findings are not.

## 1. Three copies of every outbound particle

`gatherMoversToPool` swap-pops movers **out** of the source leaves into the pool, so after the gather
the particles exist once, in `a_pool`. The exchange then adds two more copies:

- `CD_ParticleContainerImplem.H:269` — `std::vector<std::vector<char>> sendBuf(numRanks);` one byte
  buffer per destination rank (1024 of them at production scale).
- `:290-306` — each mover is `linearizeParticle`'d out of the pool into `sendBuf[r]`. **Copy 2.**
- `:340` — `std::vector<char> sflat(stot);`
- `:345` — `std::memcpy(sflat.data() + sdispl[r], sendBuf[r].data(), ...)`. **Copy 3.**

`sendBuf` is never released before `sflat` is allocated and filled, so at the moment of the memcpy
all three are resident: the pool's `ParticleSoA`, the per-rank `sendBuf`, and the flat `sflat`. The
receive side then adds `rflat` plus whatever the destination leaf's arena doubling costs during
`delinearizeAndAppend`.

### Scale

3-D `Photon` is 77 B/particle in the SoA (position 3×8 + weight 8 + `ParticleID` 8 + `RankID` 4 +
`GhostType` 1 + payload kappa and 3 velocity components 4×8).

For the configuration that crashed — ~10 boxes/rank, `max_block_size = 16`, sampling up to 512
photons/cell:

| | |
|---|---|
| cells/rank | 10 × 16³ = 40,960 |
| photons/rank at the cap | 20.97 M |
| live SoA data | **1.61 GB/rank** |
| + `growTo` doubling slack (`2 * m_capacity`) | up to 3.2 GB |
| + the three exchange copies, if most photons cross a rank | up to ~4.8 GB |

Instantaneous photon transport makes the mover fraction unusually high: `advancePhotonsInstantaneous`
calls `remap` (`CD_McPhoto.cpp:124`) and a photon jumps a full mean free path in one hop, so with
16-cell patches spread over 512 ranks most photons leave both their patch and their rank in the same
step. That is a regime drifting Ito particles never reach.

**Note:** 512/cell is a configured cap, not a measurement, and instantaneous photons are cleared each
step. The realized peak count should be confirmed before treating 4.8 GB as the operative number —
if the physics only produces tens per cell, the footprint is two orders of magnitude smaller. The
structural problem stands either way; only its severity is in question.

## 2. The exchange is capped at 2 GB, and fails silently

`:337-338`

```cpp
CH_assert(stot <= static_cast<std::int64_t>(std::numeric_limits<int>::max()));
CH_assert(rtot <= static_cast<std::int64_t>(std::numeric_limits<int>::max()));
```

`MPI_Alltoallv` takes `int` counts and displacements, so the byte totals must fit in `INT_MAX`. The
guard is `CH_assert`, **which is compiled out in `OPT` production builds** — exactly the builds that
reach this scale. Past 2 GB of outbound bytes the displacements overflow and the exchange corrupts
silently rather than aborting.

At 1.61 GB of photons per rank the margin is already thin, and it disappears entirely at 1024
photons/cell or larger boxes.

## Plan

1. **Delete `sendBuf` — pack straight into `sflat`.** The per-rank byte count is computable from the
   pool without packing: `sum over kv of (headerLen + kv.second.size() * pp)`. So do a counting pass,
   `MPI_Alltoall` the counts, allocate `sflat` once, and have the existing parallel-over-ranks loop
   write directly at `sdispl[r]`. This removes one full copy of every outbound particle **and** the
   `numRanks` small vector allocations per remap, at no extra work — the counting pass is arithmetic
   over the pool, not a second linearization. Strictly better; no trade-off to weigh.
2. **Release pool entries as they are packed.** Once `a_pool[r]`'s leaves are linearized they are
   dead. Clearing each `ParticleSoA` after packing drops the peak to one copy plus `sflat`, at the
   cost of making the pool non-const in that loop.
3. **Make the size guard unconditional.** Replace both `CH_assert`s with a real runtime check that
   aborts via `MayDay::Abort` in every build. Do this first and separately — it is a few lines and it
   converts a silent corruption into a clear failure.
4. **Then lift the 2 GB limit.** Two options:
   - *Preferred:* exchange in units of an `MPI_Type_contiguous` of `pp` bytes so counts and
     displacements are in particles rather than bytes, raising the ceiling by a factor of `pp` (77 in
     3-D). Needs the per-leaf headers padded to a multiple of `pp`, or sent in a separate small
     exchange.
   - *Fallback:* chunk the `Alltoallv` into multiple rounds when `stot` or `rtot` exceeds `INT_MAX`.
     Simpler, no layout change, costs extra collectives only in the regime that currently breaks.
5. **Add a peak-footprint diagnostic.** Report per-rank bytes in flight (pool + `sflat` + `rflat`) at
   the remap, so this is visible in a production log instead of being inferred from arithmetic. The
   max-over-ranks RSS reporting added on `unfreed-memory-breakdown` covers the outcome; this would
   cover the cause.

## Related

The `ParticleSoA` capacity high-water mark (`growTo` doubles, `shrinkToFit()` exists with zero
callers) compounds items 1–2, but it is reset by every regrid — `ParticleContainer::regrid` news
fresh holders and destroys the old arenas — so it is bounded by the regrid interval and is not
tracked as a separate defect.

rmrsk:
# Summary

Fixes the unbounded memory growth that killed a 512-core ItoKMC run with SIGKILL, removes the
duplicate send buffer from the particle exchange (#694), and makes the memory report able to see
what it could not see before.

Closes #694 

### Background

A production ItoKMC run on Betzy died by SIGKILL. Restarting from a checkpoint reproduced the same
physical state with roughly a tenth of the reported unfreed memory, which said the growth was
accumulated rather than physical.

Three separate things were in the way of finding it.

**Chombo's counter was reporting memory that did not exist.** `Vector<T>::operator=(Vector<T>&&)`
did no memory accounting, so every move-assignment of a tracked `Vector` charged its bytes and never
released them. That is a counter bug rather than a leak, but it is a counter bug in exactly the
instrument used to chase the leak. Reported separately as #693 and fixed upstream in
chombo-discharge/Chombo-3.3#24, pulled in via #695. Nothing in this PR depends on it, but no
measurement here would have been trustworthy without it.

**There was a real leak, in `AMRMultigridKrylovOp`.** `define()` marked the preconditioner scratch
uninitialised without releasing it:

```cpp
m_precondInit = false;
```

`create()` then sizes its target with `resize(n, nullptr)`, which is a no-op once the vector is
already that long. So the second and every subsequent `define()` on the same operator -- one per
regrid -- left the previous `LevelData<T>`s allocated and unreachable. The destructor was likewise
guarded on `m_precondInit`, so an operator destroyed after a redefine freed nothing at all. Cost was
`2 x (lmax + 1 - lbase)` `LevelData` objects per regrid, measured on the production run at +14
`Copier`s and +128,208 B per regrid at 1024 ranks with `lmax = 6`.

**The largest term in the footprint was invisible.** Chombo's tracking counts only what passes
through its `Arena`, i.e. `BaseFab`-backed mesh data. Particle storage never reaches it --
`ParticleSoA` allocates with `std::aligned_alloc` and the exchange buffers are plain `std::vector` --
so on a particle-heavy run `overallMemoryUsage()` was blind to the dominant consumer. On the
production run the particle arenas peaked at 789 MB on the worst rank against a 1449 MB resident set.

Separately, `ParticleContainer` built each cross-rank exchange twice: particles were linearised into
a per-rank `sendBuf` and then copied again into the flat `sflat` buffer handed to `MPI_Alltoallv`,
doubling the transient footprint of the largest allocation in a remap (#694).

### Solution

**The leak.** Release the scratch before dropping the flag, and make the destructor unconditional:

```cpp
this->clear(m_precondTmp);
this->clear(m_precondDcor);

m_precondInit = false;
```

**The exchange.** Pack straight into `sflat` and drop `sendBuf` entirely, halving the transient
footprint of the exchange. The `MPI_Alltoallv` displacement arithmetic is `int`, which caps a single
exchange at 2 GB per rank; that limit was previously enforced by `CH_assert`, which is compiled out
in `OPT` builds, and is now an unconditional `MayDay::Abort`. Pool leaves are released as soon as
they are linearised, and `gatherToPool` drains each source leaf as it bins it -- it took a `const`
source and copied, even though every caller discards the source immediately afterwards.

**The reporting.** A new `ParticleMemory` namespace tracks particle bytes, split into the arenas a
rank is holding and the payload in flight through an exchange, because a peak dominated by one calls
for a different remedy than a peak dominated by the other. `MemoryReport` gained the resident set
size from `/proc/self/status`, which is the only figure of the three that no counter can talk itself
out of: a tracked total that climbs while the resident set does not is a counter bug rather than
memory the process is holding, which is precisely the trap this investigation fell into.

The step and grid reports now print all three sources as one table rather than as thirteen labelled
lines:

```
-- ------------------------------------------------------
-- Memory (MB)              this rank  rank max  rank min
-- ------------------------------------------------------
-- Resident set               759.770   839.965   759.770
--   high-water               781.895   856.852         -
-- ------------------------------------------------------
-- Mesh arenas                664.970   732.165   664.970
--   high-water               708.247   778.731         -
-- ------------------------------------------------------
-- Particle arenas              0.214     0.225     0.192
--   high-water                 0.490     0.490         -
-- Particle buffers             0.000     0.000     0.000
--   high-water                 0.072     0.072         -
-- Particle total               0.214     0.225     0.192
--   high-water                 0.562     0.562         -
-- ------------------------------------------------------
```

Rows are quantities, the indented `high-water` row under each is the axis over *time*, and the
columns are the axis over *ranks* at this instant. Conflating those two cost real time during the
investigation. A dash marks a quantity that does not exist rather than one that is zero: the minimum
over ranks of a high-water mark is not a useful figure, because the ranks reach theirs at different
moments.

`MemoryReport` owns the whole thing: `bytesAsMB` and the table both moved there from `Driver`, which
was formatting figures it did not own, and `CD_MemoryReport.H` now includes `CD_ParticleMemory.H` so
one header reaches all three sources. Both reports call `MemoryReport::printMemoryTable(pout(),
indent)`, so they cannot drift apart.

`ParticleMemory` is a companion to Chombo's `Arena` accounting rather than an independent facility,
so it compiles in and out with `USE_MT`, and so does the table -- resident set included. Its
functions stay declared when tracking is off and their bodies vanish, which keeps the seven
allocation sites in `ParticleSoA` and `ParticleContainer` reading the same either way; a guard at
each site is exactly how an `addBytes` and its matching `removeBytes` drift apart. The guard on the
table lives in `MemoryReport` rather than at the call sites, so `Driver` carries no preprocessor
conditionals of its own.

**Documentation.** `PARTICLE_PRECISION` existed as a build variable but appeared nowhere in the
installation page, so the only way to learn about it was to read `Lib/Definitions.make`. It is now
documented under *Main settings* and in its own section: what it governs (payload columns only --
position and weight stay `double` at either setting), that it is independent of Chombo's `PRECISION`,
that both precisions fold into the configuration string, and that checkpoints are not portable
between the two, since particle records are raw bytes whose per-particle size depends on
`sizeof(ParticleReal)`.

**`ParticleSoA::swap()`.** `m_bytes` describes the arena rather than the container, so it has to
travel with `m_buffer` through every ownership transfer. `steal()` moved it and `swap()` did not,
which mattered because `catenate()` swaps whenever the destination is empty -- the common case on the
gather-all path -- and `makeSuperparticles` swaps a leaf every step. Found in review; the effect was
to understate the peak (0.490 MB against 0.388 MB on the same eight-step run).

### Side-effects

- Report format changed. Anything parsing `pout.*` for `Unfreed memory` or the `= value (MB)` lines
  will need updating. Figures are in MB to three decimals rather than in bytes, deliberately: whole
  megabytes make a ratchet of tens of kilobytes per step read as an unchanging integer.
- `ParticleContainer::gatherToPool` now takes a non-const source and empties it. Every in-tree caller
  already discarded the source; an external caller relying on it surviving would break.
- The exchange size guard now aborts in `OPT` builds where it previously did nothing.
- `ParticleMemory` counters are process-wide atomics with relaxed ordering, updated from inside
  OpenMP regions. They are diagnostic totals and impose one relaxed fetch-add per arena allocation.
- `PARTICLE_PRECISION` now rejects any value other than `DOUBLE` or `FLOAT`. It previously resolved
  with `$(if $(filter FLOAT,...),float,double)`, so every other value silently selected double --
  including `SINGLE`, which is what Chombo's own `Make.defs` comment tells you to write, and
  lower-case `float`. A build that quietly used the other precision is worse than a build that stops,
  but this does turn a previously-succeeding (and wrong) invocation into a hard error.
- Three extra `ParallelOps::min`/`max` reductions per report. These are not opt-in: both reports are
  gated on `m_verbosity > 0` rather than on `Driver.write_memory`, which controls a separate memory
  file dump. The pre-PR code reduced under the same gating, so this is a slightly larger collective
  cost in the same place rather than a new one.

### Alternative solutions

- **Trim the `ParticleSoA` arenas after a regrid** rather than counting them. Implemented and then
  reverted: `growTo` doubles, so live containers already sit at 1.0-2.0x their contents and average
  ~1.5x, and forcing a shrink to exactly 2x *raised* the measured peak (0.460 MB unmodified vs 0.546
  MB with `trimCapacity(2.0)`). The slack was never the problem -- measured against live particle
  counts the arenas carry 1.1-2.1x what the particles need. The leaf drain is the part of that work
  that survived.
- **Chunk the exchange** instead of aborting past 2 GB. Correct in the long run, and it is what the
  abort message points at, but it is a larger change than removing the duplicate buffer and is not
  needed until a rank genuinely moves that much at once.
- **Report bytes, or whole megabytes.** Bytes are unreadable in the gigabyte range; whole megabytes
  hide the signal these reports exist to carry.
- **Leave the particle counters always on.** They are cheap, but a report showing particle arenas
  next to two blank sources would invite exactly the confusion between counter and process that the
  resident-set column exists to prevent.
- **Spell the payload precision `SINGLE` or `float`.** `FLOAT` is what Chombo actually implements for
  the analogous `PRECISION` variable -- its `Make.defs.defaults` comment says `SINGLE`, but
  `Make.rules` substitutes on `FLOAT` -- and it is the only one of the three that is not lower-case
  among `DOUBLE`/`TRUE`/`FALSE`/`HIGH`.

The remaining measured growth is a transient rather than a leak, and is tracked separately: #696
(de-refinement piles `2^D` fine cells into one coarse cell before `makeSuperparticles` runs, which is
the largest single term in the 789 MB particle peak) and #697 (consolidating the container-owned
particle columns).

### PR-review checklist

Mandatory:

Optional:


rmrsk:
@claude review

rmrsk:
Thanks — both findings were real. Addressed in ad9f9dc1c.

**1. `ParticleSoA::swap()` dropping `m_bytes` — confirmed and fixed.** The reasoning holds: `m_bytes` describes the arena rather than the container, so it has to travel with `m_buffer`, and `swap()` was the one transfer that left it behind. `steal()` moved it; the three `m_buffer` assignment sites in `CD_ParticleSoAImplem.H` (`reallocate`, `permute`, the `m_size == 0` path in `shrinkToFit`) all pair `addBytes`/`removeBytes` correctly. `swap()` was the only gap, and I audited the rest of the file to confirm that.

Measured locally on an eight-step 4-rank AirBasic run: the live figure is unchanged (0.215 vs 0.216 MB), but the particle high-water goes from 0.388 MB to 0.490 MB. So the pre-fix counter was *understating* the peak — which is the wrong direction for a report whose whole purpose is to catch a peak before the scheduler does. Also expanded the `m_bytes` doc block to say it must travel with `m_buffer`, since the invariant is what was missed rather than the line.

**2. `literalinclude` drift — confirmed and fixed.** `1541-1563` → `1547-1569` (six lines, not five, once the `swap()` fix and the expanded doc block are counted). I checked every `literalinclude` pointing into a file this PR touches by rendering the excerpt from both `upstream/main` and this branch and diffing them; the nine ranges into `CD_ParticleContainer.H` are unaffected, since that change is line-neutral. `sphinx -W --keep-going` builds clean.

**3. Description gating — correct, and my write-up was wrong.** `writeMemoryTable()` is called from `gridReport()`/`stepReport()`, both gated on `m_verbosity > 0`; `m_writeMemory` controls a separate memory *file* dump. Description updated to say so rather than implying the reductions are opt-in.

rmrsk:
@claude review

rmrsk:
@claude review

PR #699, opened 2026-08-14
Carry a reduced particle across the regrid, and merge before rebuilding

### LINKED ISSUE #696: Regrid de-refinement transiently holds 2^D times the target particles per cell

rmrsk:
# Summary

Carries the bulk Ito particles across a regrid in reduced form (53 B against `ItoParticle`'s 149 B in
3-D) and runs the super-particle merge inside `ItoSolver::regrid`, so the de-refinement pile-up is
merged away before it is ever rebuilt at the larger size. Closes #696.

| commit | what |
|---|---|
| `2cebb35d` | Consolidate the per-cell merge intermediate on `ParticleManagement::MergeParticle`; delete `NonCommParticle` |
| `5afeed66` | Run every Ito merge on `ItoMergeParticle`; add `mergeLite` |
| `91fcad87` | Move the merge target and the regrid merge policy onto `ItoSolver` |
| `c2a74cdd` | Carry the bulk particles across the regrid in reduced form |
| `6389b5ee` | Repair the Sphinx `literalinclude` pins this work moved |
| `99d88cec` | Unrelated pre-existing fix — `computeSpaceChargeDensity` indexed an empty density vector |

### Background

At a de-refining regrid `2^D` fine cells collapse into one coarse cell. The particles were
redistributed onto the new grids in full `ItoParticle` form and merged afterwards in the time
stepper, so between those two points each rank held up to `2^D` times the target particles per cell
at 149 B each.

### Solution

**One pipeline for every merge method**, on `ItoMergeParticle` (position + weight + particleID +
energy):

```
extract  ->  mergeLite  ->  rebuild
```

`ItoMergeParticle` already existed in `CD_ItoSolver.cpp`, used by the `kd_*` and `nn_pair_*` merges;
it is promoted to a header and everything else moves onto it. Two particle types now span all of
merging: `ItoMergeParticle` at the container level, `MergeParticle` as per-cell AoS scratch.

**`mergeLite` owns the shape rule.** Cell-based methods (`equal_weight_kd`, `reinitialize`,
`reinitialize_bvh`, `nn_sfc`, `external`) reduce each cell to at most its per-level target and are not
followed by the container-wide split. Whole-container methods (`kd_carve`, `kd_patch`, `kd_skin_nn`,
`nn_pair_{tree,onecell,hash}`) drain over-full cells against one hierarchy-wide threshold and are.
`None` does neither — not even the split, which would otherwise refill every under-full cell at every
regrid.

**The regrid path is that pipeline with its ends detached:**

```
preRegrid()   extracts the bulk container into m_regridParticles (53 B), releasing each
              ItoParticle arena patch by patch
<regridAmr, load balancing, redistribution -- all at 53 B>
regrid()      remaps m_regridParticles, merges it, and only then rebuilds ItoParticles,
              releasing each reduced leaf as it is consumed
```

Both containers make the `preRegrid`/`regrid` round trip — the `ItoParticle` one is empty over that
window, but `ParticleContainer::regrid` is what re-defines its holders over the new layout.

**`ItoSolver` owns what to merge to and whether to merge at a regrid**, which is what lets the merge
bracket the redistribution:

```
ItoSolver.particles_per_cell      = 32       ## replaces <stepper>.particles_per_cell, BrownianWalker.ppc
ItoSolver.regrid_superparticles   = solver   ## replaces <stepper>.regrid_superparticles
```

`merge_interval` stays on `ItoKMCStepper`.

### Side-effects

**Behavioural**

- **The pack/unpack round-trip runs at a regrid even with `regrid_superparticles = none`**, zeroing
  `mobility`, `diffusion`, `vx/vy/vz`, `scratch*` and `old_*`. Every stepper recomputes those after a
  regrid, and any merge already discarded them. `particleID` is carried explicitly. One consequence
  is persisted: `h5Columns` is the velocity triple, so a checkpoint taken across a regrid stores
  zeros there — `postCheckpointSetup` recomputes them before use.
- **Godunov's regrid merge moves before its Poisson solve**, matching the base `ItoKMCStepper`. Only
  observable at `m_timeStep == 0`; a merge conserves total weight.
- **`ItoSolver::initialData()` seeds at `particles_per_cell`, not `ppc_restart`** — a cold start is
  not a restart. This changes the seeded population wherever the two differ: four of the six in-tree
  ItoKMC inputs (32 vs 16) and the 3-D BrownianWalker cases.
- **BrownianWalker's merge target changes hands** and it gains a regrid-time merge in place of the
  second merge it ran after `m_solver->regrid`.

**API and input breaks**

- `ItoSolver::setParticleCellMerger` takes a `ParticleMerger<ItoMergeParticle>`.
- The eleven public `makeSuperparticlesXxx(WhichContainer, ...)` entry points are gone; the extract
  and rebuild ends now live in the dispatcher. `makeSuperparticles(container, ppc, method)` still
  reaches any single method by name.
- `CD_NonCommParticle.H` is deleted — a public type under `Source/Particle`, so out-of-tree users of
  `NonCommParticle<M,N>` break. Nothing in-tree uses it after the first commit.
- Two keys removed from each ItoKMC stepper prefix, `BrownianWalker.ppc` removed, two `ItoSolver.`
  keys added. 28 `.inputs` and 4 `.options` files change; out-of-tree inputs abort at startup on the
  missing mandatory keys.
- `parseParticleMerger` aborts when a per-level `particles_per_cell` is combined with a
  whole-container method, which reads entry 0 only.
- `ItoKMCGodunovStepper`'s timer loses its "Make superparticles" line for the regrid; the `advance()`
  line stays and the new seam keeps its own `CH_TIME`.

**Load-balance ordering.** `BrownianWalkerStepper::loadBalanceBoxesParticles` runs between
`preRegrid` and `regrid`, where the `ItoParticle` container is now empty, and is retargeted at the
reduced holder via `ItoSolver::getRegridParticles()`. An all-zero load vector produces a valid but
unbalanced layout with nothing to signal it. The two sibling balancers that are safe only by their
ordering relative to `preRegrid` are commented as such.

**`99d88cec` is unrelated to this work and can be split out.**
`ItoKMCStepper::computeSpaceChargeDensity` indexed element 0 of the Ito and CDR density vectors in
its realm asserts. A chemistry that puts every plasma species on the Ito solver leaves the CDR vector
empty, so any `DEBUG` build of such a configuration aborted before its first time step.

### Alternative solutions 

Giving `MergeParticle` `weight()`/`position()` accessors instead of switching the `ParticleManagement`
template parameters to member-data pointers; keeping the eleven per-method entry points as
`WhichContainer` wrappers; bounding the `2^D` spike itself by chunking `ParticleContainer::regrid`
(out of scope — merging does not belong in `ParticleContainer`, which also serves photons and tracer
particles).

### Verification

- Particle counts **byte-identical to `main`** across all 11 BrownianWalker merge-method regression
  tests, 1000 steps each, for commits 1, 2 and 4. Commit 3 shows a bounded ±1.4 % drift diverging at
  the first regrid — BrownianWalker temporarily without a regrid merge — which commit 4 restores.
- With every merge off, a regrid changes the particle count by nothing: the per-step deltas at regrid
  steps are indistinguishable from those around them. This is the check that `mergeLite` does not
  split when it should not.
- All 11 methods at `OPT=TRUE DEBUG=TRUE` in 2-D; `kd_carve` also in 3-D, on 4 MPI ranks, and with
  `OMP_NUM_THREADS=4`. `external` with the default no-op merger does not merge, confirming the user
  hook reaches `applyCellMerger` on the reduced container.
- `verify_conservation` passes; both load-balance modes give non-degenerate box counts; a checkpoint
  taken across a regrid restarts cleanly.
- `pre-commit run --all-files` and `sphinx -W --keep-going` pass. Every moved `literalinclude` pin was
  re-rendered and word-diffed against its pre-change excerpt — one had drifted silently.

**Memory.** On a de-refining 2-D BrownianWalker case, peak particle arenas fall 2236 MB → 1596 MB and
peak RSS 2669 MB → 1751 MB, with the final particle count unchanged. Splitting that run shows both
sides peak during initial data rather than at a regrid, so those figures measure the arena discipline
added to extract/rebuild, not the regrid spike. Neither in-tree case can isolate the spike on one rank
(BrownianWalker merges every step; the ItoKMC JSON case has `max_amr_depth = 0`), so the
production-shaped measurement is still outstanding.

### PR-review checklist

Mandatory:

Optional:


rmrsk:
@claude review

PR #704, opened 2026-08-20
Compare CdrPlasma and ItoKMC on the same problem, and fix the defects it exposed

### LINKED ISSUE #701: KMCSolver::advanceTau never applies the tau-leaping condition
## Bug

`KMCSolver::advanceTau` (`Source/KineticMonteCarlo/CD_KMCSolverImplem.H:848`) selects its substep as

```cpp
curDt = a_dt - curTime;
```

and then refines it only when the resulting state comes out invalid:

```cpp
valid = state.isValidState();

if (valid) { a_state = state; curTime += curDt; }
else       { curDt *= 0.5; }
```

The tau-leaping condition `tau = eps * X_i / |sum_j nu_ij a_j|` is never applied. The full interval
handed to `advanceTau` is therefore integrated in a single leap unless a population goes negative,
and the accuracy of that leap is uncontrolled.

Two consequences:

1. `ItoKMCJSON.prop_eps` has no effect on any propagator that routes through `advanceTau`, i.e. the
   `explicit_euler`, `midpoint`, `prc` and `implicit_euler` selectors. The value does reach
   `KMCSolver::m_eps` via `setSolverParameters`, but `m_eps` is read at exactly one site
   (`CD_KMCSolverImplem.H:353`, inside `getNonCriticalTimeStep`), and `getNonCriticalTimeStep` is
   only ever called from `advanceHybrid`. The `hybrid_*` selectors are unaffected.

2. Where the leap parameter `u = a*dt/X` is O(1), the propagators lose their formal order and
   systematically under-produce a growing species.

`advanceHybrid` already does this correctly (`CD_KMCSolverImplem.H:990`):

```cpp
Real dtNonCrit = this->getNonCriticalTimeStep(a_state, nonCriticalReactions, propensitiesNonCrit);

const Real curDt = std::min(a_dt - curTime, std::min(dtCrit, dtNonCrit));
```

## Fix

Bound the substep by the leap condition in the outer loop of `advanceTau`:

```cpp
const std::vector<Real> propensities = this->propensities(a_state, a_reactions);

const Real dtLeap = this->getNonCriticalTimeStep(a_state, a_reactions, propensities);

curDt = std::min(a_dt - curTime, dtLeap);
```

This makes `prop_eps` live for every tau propagator and restores the intended accuracy control. The
inner validity loop can remain as a backstop.

### LINKED ISSUE #702: CdrPlasmaJSON photon production is not accumulated across chemistry substeps
## Bug

`CdrPlasmaJSON::integrateReactions` (`Physics/CdrPlasma/PlasmaModels/CdrPlasmaJSON/CD_CdrPlasmaJSON.cpp:5823`)
sub-cycles the reactive advance whenever the transport step exceeds `CdrPlasmaJSON.chemistry_dt`:

```cpp
const int numSteps = std::ceil(a_dt / m_chemistryDt);

for (int step = 0; step < numSteps; step++) {
  const Real dt = a_dt / numSteps;

  ... integrateReactionsExplicitRK2(a_cdrDensities, a_photonProduction, ..., dt, ...);
}
```

The plasma densities accumulate across the substeps, because each integrator updates them in place:

```cpp
a_cdrDensities[i] = a_cdrDensities[i] + a_dt * (c1 * cdrK1[i] + c2 * cdrK2[i]);
```

The photon production does not. All three integrators **assign** to it rather than accumulating:

* `integrateReactionsExplicitEuler`, line 6081: `a_photonProduction[i] = rteSources[i] * a_dt;`
* `integrateReactionsExplicitRK2`, line 6126: `a_photonProduction[i] = a_dt * (c1 * rteK1[i] + c2 * rteK2[i]);`
* `integrateReactionsExplicitRK4`, line 6193: `a_photonProduction[i] = a_dt * (rteK1[i] + 2.0 * rteK2[i] + 2.0 * rteK3[i] + rteK4[i]) / 6.0;`

so after the loop only the final substep's contribution survives, over the substep interval `a_dt/numSteps`
rather than over `a_dt`. The photon production handed to the radiative transfer solver is therefore too
small by a factor of `numSteps`.

The error is silent and its size is set by an input parameter: `numSteps = ceil(a_dt / chemistry_dt)`.
It is absent at the default `chemistry_dt = 1.E99`, where `numSteps == 1`, and grows as `chemistry_dt` is
lowered to resolve the chemistry. Lowering `chemistry_dt` therefore improves the density integration and
degrades the photon production at the same time.

## Fix

`photonProduction` is already zero-initialised by the caller
(`CD_CdrPlasmaJSON.cpp:5235`, `std::vector<Real> photonProduction(m_numRtSpecies, 0.0);`), so accumulating
in the integrators is sufficient and needs no other change:

```cpp
a_photonProduction[i] += rteSources[i] * a_dt;                                       // line 6081
a_photonProduction[i] += a_dt * (c1 * rteK1[i] + c2 * rteK2[i]);                     // line 6126
a_photonProduction[i] += a_dt * (rteK1[i] + 2.0 * rteK2[i] + 2.0 * rteK3[i] + rteK4[i]) / 6.0;  // line 6193
```

Note that the integrators are also callable directly, not only through `integrateReactions`; any such
caller must zero the array before the call for `+=` to be correct.

rmrsk:
# Summary

Adds `Exec/Examples/ItoKMC/ComparisonCdrPlasma`, which runs the same streamer problem through `CdrPlasma`
and `ItoKMC`, and fixes the defects that comparison exposed.

Closes #701. Closes #702.

# Background

The example runs a rod-needle electrode at 35 kV in a grounded 8 cm x 8 cm box of atmospheric air with both
plasma models, selected by a single switch. Geometry, mesh, voltage, transport data, chemistry, and initial
seed are shared, so the models can be compared directly.

They did not agree. Driving the disagreement to zero required routing every `ItoKMC` species through a CDR
solver, so that both models integrate the same equations with the same `CdrCTU` code, and then removing
differences one at a time. Six turned out to be defects rather than modelling choices. With all six fixed
the two models agree on the streamer head field to 0.02%, and on the radius and velocity as well.

# Solution

**`CdrPlasmaJSON` discarded photon production across chemistry substeps (#702).**
`integrateReactions` sub-cycles the reactive advance when the transport step exceeds
`CdrPlasmaJSON.chemistry_dt`. Densities accumulate across substeps but the photon production was assigned
rather than accumulated, so only the final substep survived and the production handed to the radiative
transfer solver was too small by `ceil(dt/chemistry_dt)`. The error is absent at the default
`chemistry_dt = 1.E99` and grows as the chemistry is resolved, so lowering `chemistry_dt` improved the
density integration and degraded the photon production at the same time.

**`KMCSolver::advanceTau` never applied the tau-leaping condition (#701).**
The substep was bounded only by the remaining splitting interval and refined only when the resulting state
came out invalid, so `tau = eps*X_i/|sum_j nu_ij a_j|` was never applied and `ItoKMCJSON.prop_eps` had no
effect on the `explicit_euler`, `midpoint`, `prc` and `implicit_euler` selectors. `advanceHybrid` already
applied the bound correctly. Verified against `Exec/Convergence/KineticMonteCarlo/C1`, whose mean is
`N0*exp(t)`: over `t = 20` with `N0 = 2e7` the exact answer is 9.70e15 and the pre-fix answer was 2.59e10;
afterwards the error falls monotonically with `eps` and the propagators recover their formal order.

**`McPhoto::regrid` zeroed the absorbed-photon density.**
`m_phi` holds the density of the photons absorbed during the previous advance, which is the population in
`m_bulkPhotons`. `regrid` re-deposited `m_photons` instead -- the in-flight container, which
`advancePhotonsInstantaneous` has already drained -- and since deposition resets its target, `m_phi` came
back zero. `CdrPlasma` reads that field one step later in `addPhotoIonization`, so the step following every
regrid had no photoionization at all, with an error that scaled with regrid frequency.

**`ItoKMCGodunovStepper` passed half the time step to the CTU predictor.**
`CdrCTU::upwind` reconstructs as `phi + 0.5*(1 - v*a_extrapDt/dx)*slope` with undivided slopes, so the
argument is twice the time extrapolated over and centring the flux at the half step requires the full `dt`.
Passing `0.5*dt` centred at `t + dt/4` and halved the transverse CTU correction. Every other caller in the
tree was audited against this contract and found correct; the `a_extrapDt` doxygen, which described the
parameter as "shifting of div(F) towards e.g. half time step", is corrected.

**`CdrPlasmaGodunovStepper` dropped the diffusion term from the semi-implicit space charge.**
The `dt*Z*Qe*div(D*grad(phi))` contribution to rho-dagger had been disabled behind `#if 0` since #150, while
`ItoKMC` includes it. It is worth about 1% of the local space charge, concentrated at the streamer head. The
adjacent comment also had the sign backwards.

**`ItoKMCGodunovStepper::regrid` built the CDR conductivity from uninitialized memory.**
It called `computeConductivities()`, which rebuilds the CDR contribution from `m_cdrMobilities` -- but
`allocateInternals()`, a few lines earlier in the same routine, has just reallocated those, and nothing
fills them until `computeDriftVelocities()` at the end of regrid. Reordering cannot fix it, since
`computeMobilities()` needs a field that is only valid after the solve being set up. The class already had
`m_semiImplicitConductivityCDR` for this, with a scratch holder, an `interpToNewGrids` and checkpoint
read/write -- only the fill was missing. The conductivity is now carried across a regrid as a mesh variable,
as the space charge already was, so a regrid on restart follows the same path as any other regrid.

Also in this PR: the semi-implicit CDR buffers are given a defined value at allocation, since `regrid` may
consume them before an advance has filled them; the cell conductivity is floored, as `CdrPlasma` floors its
own; the regrid interpolation of the semi-implicit data uses slope-limited rather than piecewise-constant
interpolation, matching `CdrPlasma` (see #705); and `CD_ItoKMCJSON.options` no longer advertises
`tau_plain`, `tau_midpoint` and `hybrid_plain`, none of which `parseAlgorithm` accepts.

# Side-effects

* **#702 changes `CdrPlasma` results whenever `chemistry_dt` is smaller than the transport step.** Photon
  production, and hence photoionization, increases by up to `ceil(dt/chemistry_dt)`. Runs at the default
  `chemistry_dt = 1.E99` are unchanged. Benchmarks produced with a lowered `chemistry_dt` will move.
* **#701 changes the `explicit_euler`, `midpoint`, `prc` and `implicit_euler` selectors**, which now substep
  rather than leaping the whole interval. The `hybrid_*` selectors are untouched.
* **The `McPhoto` fix changes any `CdrPlasma` + `McPhoto` run with AMR**, which is the normal configuration.
  Photoionization is no longer lost on the step after each regrid, so results move by an amount that scales
  with regrid frequency.
* The CTU and rho-dagger fixes change `ItoKMC` and `CdrPlasma` results respectively in any run that uses
  them.
* Checkpoints written before this PR contain zeros for `m_semiImplicitConductivityCDR`, so restarting from
  one and regridding gives a zero CDR conductivity on the first regrid. Given the fixes above, pre-PR
  checkpoints are not comparable to post-PR runs regardless.
* The `m_dirtySampling` branch of `McPhoto` has the same photon-density defect and cannot be repaired by
  re-deposition, since it writes `a_phi` through a local container that is discarded. It is off by default
  and is left for separate work.

# Alternative solutions

* **#702** could zero `a_photonProduction` inside each integrator rather than relying on the caller.
  Rejected: `integrateReactions` must accumulate across its substeps, so that would reintroduce the bug. The
  requirement is documented on the parameter instead.
* **#701** could apply the leap bound inside each of the four propagators, or partition the reactions into
  critical and non-critical as `advanceHybrid` does. The first duplicates the bound four times; the second
  turns `advanceTau` into `advanceHybrid`.
* **The CDR conductivity** could be recovered at regrid by rebuilding the electric field from the regridded
  potential and recomputing the mobilities, or by carrying the mobilities themselves across the regrid.
  Both recompute what can simply be stored, and the second carries one field per CDR species where one field
  suffices.
* **The diffusive flux** transported by the Ito solver differs from the CDR solver's wherever the diffusion
  coefficient varies in space. That is a modelling question rather than a defect and is not addressed here;
  it is reported as #706.

# PR-review checklist

Mandatory:

Optional:


rmrsk:
@claude review

rmrsk:
@claude review

PR #709, opened 2026-08-20
Evaluate the ItoKMC chemistry at a selectable interpolant between the start- and end-of-step fields

### LINKED ISSUE #707: ItoKMC chemistry should be evaluated at a selectable interpolant between the fields at the start and end of the step
## Summary

`ItoKMCGodunovStepper` evaluates the chemistry at the electric field from the *end* of the transport step.
In the semi-implicit formulation that is the already-screened field, so when the field is changing during a
step the reaction rates are systematically biased. The chemistry should instead be evaluated at a
user-selectable interpolant between the fields at the start and end of the step, defaulting to the midpoint.

## Detail

`ItoKMCGodunovStepper::advance` calls `advanceReactionNetwork(a_dt)`, which forwards to
`ItoKMCStepper::advanceReactionNetwork(m_electricFieldFluid, a_dt)`. By that point
`m_electricFieldFluid` holds `E^(k+1)` -- the field produced by the semi-implicit Poisson solve inside
`advanceEulerMaruyama`. The whole reactive substep therefore runs at the end-of-step field.

Where the plasma is screening the field, `|E^(k+1)| < |E^k|` and the chemistry sees a field that is too low
for the entire step. The error is first order in `dt` and is a bias rather than a random error, so it does
not average out over many steps: it systematically under-predicts ionization.

The bias is amplified by the sensitivity of the Townsend coefficient to the field. From the BOLSIG+ table
used by the examples:

| E/N | `dln(alpha)/dln(E)` |
|---|---|
| 200 Td | 3.8 |
| 300 Td | 2.9 |
| 450 Td | 2.2 |
| 600 Td | 1.8 |

so a 10% field error becomes a 20-40% error in the ionization rate, and it is worst at the *lower* fields
ahead of the streamer head rather than at the peak. For scale, the fractional field change within a step
under the semi-implicit response is `(dt/tau_M)/(1 + dt/tau_M)`, where `tau_M = eps0/sigma` is the
dielectric relaxation time -- roughly 13 ps at `n_e = 1e20 m^-3` and 1.4 ps at `1e21 m^-3` in atmospheric
air, against typical streamer time steps of a few ps.

## Proposed change

Evaluate the chemistry at

```
E_reactive = (1 - theta) * E^k + theta * E^(k+1)
```

with `theta` set by a new input option:

```
ItoKMCGodunovStepper.reactive_E_centering = 0.5
```

* Default `0.5`, i.e. the time-centred field, which removes the leading-order bias.
* `theta = 1` reproduces the present behaviour, so the change can be backed out from the input file.
* `theta = 0` evaluates at the field the step started from.
* **Values outside `[0, 1]` must abort at run time.** Extrapolating outside the interval has no physical
  justification and would silently produce fields that were never solved for.

## Implementation notes

The plumbing already exists -- `ItoKMCStepper::advanceReactionNetwork(const EBAMRCellData& a_E, const Real
a_dt)` takes the field explicitly and threads it down to the per-cell kernels, asserting only that it lives
on the fluid realm. So the change is confined to `ItoKMCGodunovStepper`:

1. Add an `EBAMRCellData` member on the fluid realm, allocated in `allocateInternals`.
2. Copy `m_electricFieldFluid` into it at the top of `advance`, before the transport step, while it still
   holds `E^k`.
3. After the transport step, form the interpolant with `m_electricFieldFluid`, which now holds `E^(k+1)`.
4. Pass it to the explicit-field overload of `advanceReactionNetwork`.
5. Parse and validate `reactive_E_centering` alongside the other options, aborting outside `[0, 1]`.

A working prototype of the `theta = 0.5` case exists and is a handful of lines.

## What this must NOT be extended to

The drift update must keep using `E^(k+1)` with the lagged mobility `mu^k`. That pairing is not a modelling
choice -- it is what the semi-implicit derivation assumes:

```
rho^(k+1) = rho^dagger - dt*div(sigma^k E^(k+1))
  =>  div((eps + sigma^k*dt) E^(k+1)) = rho^dagger
```

which is precisely the operator `setupSemiImplicitPoisson` builds. Advecting with a centred field while the
operator still carries `sigma*dt` would mean the charge actually moved no longer matches what the solve
assumed, and `E^(k+1)` would stop being the solution of the coupled system -- giving up the implicitness
that removes the dielectric-relaxation constraint.

A consistent centred drift is derivable, but it is a different scheme: `eps + sigma*dt/2` in the operator
plus an explicit `-(dt/2)*div(sigma^k E^k)` on the right-hand side, i.e. Crank-Nicolson rather than
backward Euler. That is second order only while `dt <~ tau_M`; beyond it the response stops damping and
starts oscillating -- the amplification factor `(1 - r/2)/(1 + r/2)` reaches -0.39 at `r = dt/tau_M = 4.6`
and tends to -1 for large `r`, where backward Euler gives a clean monotone 0.18. Since `dt >> tau_M` is
exactly the regime the semi-implicit scheme exists to operate in, the present backward-Euler drift should
stay as it is.

The reactions are safe to centre precisely because they sit *outside* the semi-implicit coupling: nothing in
the Poisson operator makes any assumption about which field the rate coefficients are evaluated at.

rmrsk:
# Summary

Resolves #707.

`ItoKMCGodunovStepper` evaluated the chemistry at the electric field from the *end* of the transport step. In the semi-implicit formulation that is the already-screened field, so the reaction rates were systematically biased low wherever the plasma screens the field. The chemistry is now evaluated at a user-selectable interpolant between the fields at the start and end of the step, defaulting to the midpoint:

```
ItoKMCGodunovStepper.reactive_E_centering = 0.5
```

### Background

`ItoKMCGodunovStepper::advance` called `advanceReactionNetwork(a_dt)`, which forwards to `ItoKMCStepper::advanceReactionNetwork(m_electricFieldFluid, a_dt)`. By that point `m_electricFieldFluid` holds `E^(k+1)` — the field produced by the semi-implicit Poisson solve inside `advanceEulerMaruyama` — so the whole reactive substep ran at the end-of-step field.

Where the plasma is screening the field, `|E^(k+1)| < |E^k|` and the chemistry sees a field that is too low for the entire step. The error is first order in `dt` and is a *bias* rather than a random error, so it does not average out over many steps: it systematically under-predicts ionization.

The bias is amplified by the sensitivity of the Townsend coefficient to the field. From the BOLSIG+ table used by the examples, `dln(alpha)/dln(E)` runs from 3.8 at 200 Td to 1.8 at 600 Td, so a 10% field error becomes a 20–40% error in the ionization rate — and it is worst at the *lower* fields ahead of the streamer head rather than at the peak.

### Solution

The chemistry is evaluated at

```
E_reactive = (1 - theta) * E^k + theta * E^(k+1)
```

with `theta` set by the new input option `ItoKMCGodunovStepper.reactive_E_centering`. The change is confined to `ItoKMCGodunovStepper`, because `ItoKMCStepper::advanceReactionNetwork(const EBAMRCellData& a_E, const Real a_dt)` already takes the field explicitly and threads it down to the per-cell kernels:

1. A new `EBAMRCellData m_reactiveElectricField` on the fluid realm, allocated in `allocateInternals`.
2. `advance` copies `m_electricFieldFluid` into it at the top, before the transport step, while it still holds `E^k`.
3. After the transport step the interpolant is formed in place against `m_electricFieldFluid`, which now holds `E^(k+1)`.
4. It is handed to the explicit-field overload of `advanceReactionNetwork`.
5. `parseReactiveFieldCentering` parses the option in both `parseOptions` and `parseRuntimeOptions`, and aborts outside `[0, 1]`.

Defaults and limits:

* Default `0.5`, i.e. the time-centred field, which removes the leading-order bias.
* `theta = 1` reproduces the previous behaviour, so the change can be backed out from the input file.
* `theta = 0` evaluates at the field the step started from.
* Values outside `[0, 1]` abort at run time. Extrapolating outside the interval has no physical justification and would silently produce fields that were never solved for.

The centring applies to the reactions only. The drift update keeps using `E^(k+1)` with the lagged mobility `mu^k`, because that pairing is what `setupSemiImplicitPoisson` builds its operator for:

```
rho^(k+1) = rho^dagger - dt*div(sigma^k E^(k+1))
  =>  div((eps + sigma^k*dt) E^(k+1)) = rho^dagger
```

Advecting with a centred field while the operator still carries `sigma*dt` would mean the charge actually moved no longer matches what the solve assumed, and `E^(k+1)` would stop being the solution of the coupled system. The reactions are safe to centre precisely because they sit *outside* the semi-implicit coupling: nothing in the Poisson operator makes any assumption about which field the rate coefficients are evaluated at.

**Verification**

* Builds clean in 2-D and 3-D at `OPT=HIGH`, and in a `DEBUG=TRUE OPT=TRUE` build so `CH_assert` is live (including the realm assertion inside `advanceReactionNetwork`). Both ItoKMC regression tests run to completion; the assertion build ran 50 steps with no trips.
* `reactive_E_centering = 1.5` and `= -0.1` abort at parse time with the intended message.
* **`theta = 1` backs the change out exactly.** A baseline binary was built from the pre-change tree and run against the modified binary at `theta = 1` on the AirBasic example (2-D, 300 steps, 12 ranks). The per-step `(time, Emax, max n/N, Qplus)` series are identical over all 300 steps.
* **`theta = 0.5` moves the answer in the predicted direction.** Same setup: at matched physical time (`t ~ 1.08e-8 s`) `Qplus` is `4.30e-6` against `3.17e-6`, i.e. ~36% more ionization — inside the 20–40% band the sensitivity argument predicts.
* `pre-commit run` passes all hooks on the changed files, and `sphinx-build -W --keep-going` builds with no new warnings.

### Side-effects

* **The default changes the answer for every existing ItoKMC input deck.** Going from an effective `theta = 1` to `theta = 0.5` is not a small perturbation — see the ~36% `Qplus` difference above. Both ItoKMC regression tests will produce different output than on `main`. Setting `reactive_E_centering = 1` restores the old results bit-for-bit.
* One additional `SpaceDim`-component `EBAMRCellData` is allocated on the fluid realm. It is reallocated by `allocateInternals` on every regrid and is not carried across one, since it is written and consumed entirely within a single `advance`.
* The option was added to `CD_ItoKMCBackgroundEvaluator.options` as well. `ItoKMCBackgroundEvaluator` derives from `ItoKMCGodunovStepper` and re-parses under its own name, so a mandatory `pp.get` would otherwise abort it.
* The option was added to the 7 `.inputs` files that carry an `ItoKMCGodunovStepper` block, since the codebase requires inputs to be mandatory rather than queried.
* A new `Reactive field centering` section was added to `Docs/Sphinx/source/Applications/ItoKMC.rst`. The new option was appended at the *end* of the `.options` file so that the existing `literalinclude` with `:lines: 19-31` is unaffected.

### Alternative solutions 

* **Leave the reactive field at `E^(k+1)`** — i.e. do nothing. Rejected: the bias is first order in `dt`, does not average out, and is amplified 2–4x by the field sensitivity of the Townsend coefficient.
* **Centre the drift as well (Crank-Nicolson).** A consistent centred drift is derivable — `eps + sigma*dt/2` in the operator plus an explicit `-(dt/2)*div(sigma^k E^k)` on the right-hand side — but it is a different scheme, and second order only while `dt <~ tau_M`. Beyond that the response stops damping and starts oscillating: the amplification factor `(1 - r/2)/(1 + r/2)` reaches `-0.39` at `r = dt/tau_M = 4.6` and tends to `-1` for large `r`, where backward Euler gives a clean monotone `0.18`. Since `dt >> tau_M` is exactly the regime the semi-implicit scheme exists to operate in, the backward-Euler drift is left as it is.
* **Hard-code `theta = 0.5` with no input option.** Rejected: an option makes the change reversible from the input file, which is what lets existing results be reproduced and lets the two choices be compared on the same build.
* **Reuse the existing `m_fluidScratchD` instead of adding a member.** `m_fluidScratchD` already exists on the fluid realm with `SpaceDim` components and is currently unused, so it was considered first. Rejected: it is a `mutable` base-class scratch holder, while this data has to survive the entire transport step (Poisson solves, velocity and diffusion recomputation). Parking long-lived state in a scratch that any base-class routine may legitimately claim is a latent bug. The new member is the same container type, with an owner.

### PR-review checklist

Mandatory:

Optional:


rmrsk:
@claude review

Issue #710, opened 2026-08-26
AMR exchange blocks once per level; the Copier's per-rank coalescing can be fused across the whole hierarchy


rmrsk:
## Summary

`EBAMRData<T>::exchange()` runs one complete MPI round per AMR level, so an AMR-wide exchange
costs **2L blocking `MPI_Waitall`s** (4L for `preAllocatable() == 2` types) and posts
**Σ_ℓ N_ℓ** messages, where N_ℓ is the number of distinct neighbour ranks on level ℓ. Serialised
this way, each level's load-imbalance skew *adds* rather than overlapping.

Chombo's `Copier` already coalesces all traffic bound for a given rank into one `MPI_Isend`, but
only within a single Copier — hence only within a single level. The mechanism generalises: sort
the buffer across all levels with `procID` first and `level` second, and the existing merge loop
collapses the hierarchy into **one message per neighbour rank and one wait**.

The pieces needed to build this are all public Chombo API, so no changes to `Copier` or
`BoxLayoutData` are required. Nothing here has been measured yet.

§5 applies the same analysis to `AmrMesh::interpGhost` / `interpGhostPwl` / `interpGhostMG`. They
carry the same per-level blocking pattern with a *second* serialisation nested inside it — a
blocking `copyTo` per component.

**The plan is phased by blast radius, not by size of win.** Phase A is local rewrites in this
repository that leave results bit-identical and do not touch communication structure; phase B
reorders existing MPI calls; only phase C writes new communication code. Phase A lands in full
before phase B starts.

## Checklist

**Phase A — bit-identical local rewrites, chombo-discharge only**

- [x] **A1** — widen the coarsened-fine buffer to `a_variables.size()` components and hoist the `copyTo` out of the component loop, in both interpolators
- [x] **A2** — remove the redundant per-level exchange in the PWL path, behind a `bool a_doExchange = true`
- [x] **A3** — hoist the coarsened-fine buffer to a member, sized once per regrid
- [x] **A4** — reuse the MF alias shells instead of `new`-ing 2(L+1) `LevelData<EBCellFAB>` per call
- [x] **A5** — carry the four `PhaseRealm` masks as one 4-component `LevelData`

**Measurement — Chombo-side, independent of phase A**

[...]

PR #711, opened 2026-08-30
Cut the redundant MPI round trips out of ghost-cell interpolation

rmrsk:
Phase A of #710.

# Summary

### Background

`AmrMesh::interpGhost` / `interpGhostPwl` / `interpGhostMG` fill the fine-grid ghost cells across each
coarse-fine interface by walking the AMR hierarchy finest-to-coarsest. Every level in that loop does a
blocking `copyTo` to pull the coarse data onto the coarsened fine grids, and the hierarchy is exchanged
afterwards. That gives one blocking `MPI_Waitall` pair per level, which is the pattern #710 is about.

Two things made it worse than the level count alone implies:

1. Both interpolators sized their coarsened-fine buffer with `nComp = 1`
   (`CD_EBGhostCellInterpolator.cpp:183`, `CD_EBLeastSquaresMultigridInterpolator.cpp:209`). Interpolating a
   `C`-component field therefore ran a separate blocking `copyTo` per component, i.e. a second serialisation
   nested inside the level loop. The buffer itself was constructed inside `interpolate` /
   `coarseFineInterp`, so every call rebuilt an `EBCellFactory` and reallocated the whole coarsened-fine
   layout.
2. `EBGhostCellInterpolator::interpolate` ends with `a_phiFine.exchange(a_variables)`, and
   `interpGhostPwl` then exchanges every level again — a strict superset, since the trailing call also
   covers level 0. Nothing consumes the per-level exchange in between: `defineBuffers` builds `m_copier`
   from the coarse level's *valid* region only, so level ℓ never reads level ℓ−1's ghost cells, and the
   loop is descending, so no level is revisited after its own exchange.

Blocking `MPI_Waitall` counts for `L` levels and `C` components, before and after:

| Call | Before | After |
|---|---|---|
| `interpGhostPwl(EBAMRCellData, …)` | 2[(L−1)(C+1) + L] | 2[(L−1) + L] |
| `interpGhostMG(EBAMRCellData, …)` | 2[(L−1)C + L] | 2[(L−1) + L] |
| the `MFAMRCellData` variants | double the above | double the above |

For the `SpaceDim`-component gradient interpolated at `CD_CdrPlasmaStepper.cpp:498` in 3-D with L=5, that
is 42 waits down to 18.

This is the first tier of #710 by blast radius, not by size of win: it is local to this repository, leaves
results bit-identical, and does not change which ranks talk to each other or in what order. Nothing in it
depends on the `FusedCopier` work in phases B and C.

### Solution

Covers **A1**, **A2** and **A3** from the #710 checklist.

**A1 — one round trip per call instead of one per component.** Both interpolators now size the
coarsened-fine buffer to `a_variables.size()` and issue a single
`a_phiCoar.copyTo(a_variables, phiCoFi, Interval(0, numVars - 1), m_copier)` outside the component loop.
The per-component work that follows reads component `icomp - a_variables.begin()` of the buffer instead of
component 0; in the MG path that is also the source component handed to
`m_aggCoarStencils[din]->apply`, which was previously the hardwired `m_comp`.

**A2 — no duplicate exchange in the PWL path.** `EBGhostCellInterpolator::interpolate` takes a new
`bool a_doExchange = true`. `AmrMesh::interpGhostPwl(EBAMRCellData, …)` passes `false` because it exchanges
the whole hierarchy right after the loop. `AmrMesh::interpGhost(LevelData<EBCellFAB>&, …)` calls
`interpolate` directly and has no trailing exchange of its own, so it keeps the default and is unaffected.

**A3 — the buffer is a member, sized once per regrid.** `m_coarsenedFineData` is defined in
`defineBuffers()` rather than constructed per call. It holds `SpaceDim` components, which covers every
caller on the hot path: velocities, electric fields and fluxes are `SpaceDim`-component, densities and
potentials are scalars. A call that asks for more takes a local buffer for the duration of the call. Both
paths still fetch every variable in a single `copyTo`, so the round-trip count is the same either way.

A hard cap on the member was not an option: `CdrPlasmaStepper::computePhysicsPlotVars` runs the physics
plot variables through `interpGhostPwl`, and their count comes from the user's plasma model, so it exceeds
`SpaceDim` for most chemistries. Sizing the member to the widest `Interval` ever requested was not either —
the interpolators are shared per realm, phase and level, so that is a max over every caller for the rest of
the regrid rather than a current, and one wide plot-variable call would pin it there.

**Drive-by.** The three `MFAMRCellData` ghost-interpolation variants had their own per-level
`MultifluidAlias::aliasMF` loop, which is line-for-line what `AmrMesh::alias(EBAMRCellData, phase,
MFAMRCellData)` already does a few hundred lines up in the same file. They now call `allocatePointer` and
`alias` instead. This also fixes an allocation each of them was doing unconditionally: they `new`-ed
`2(L+1)` `LevelData<EBCellFAB>` shells per call and then populated only the phases that have an EB index
space, so a gas-only run allocated half of them for nothing.

### Side-effects

- **Memory.** Each `EBGhostCellInterpolator` and `EBLeastSquaresMultigridInterpolator` now holds a resident
  `SpaceDim`-component coarsened-fine `LevelData<EBCellFAB>` from `defineBuffers()` until the next regrid,
  where before the storage was transient. The coarsened-fine layout is `1/2^SpaceDim` of the fine level per
  component, per level, per phase, per realm. In exchange, the per-call `EBCellFactory` construction and
  layout allocation are gone from the hot path.
- **API.** `EBGhostCellInterpolator::interpolate` gains a defaulted trailing argument. Existing callers
  compile and behave unchanged.
- **Results.** Bit-identical. The interpolation stencils, the data that reaches them, and the order in
  which ranks communicate are all unchanged; only the number of round trips carrying that data differs.
- **Timers.** `EBGhostCellInterpolator::interpolate` reports `buffer_define`, `copy_exchange`,
  `interp_cells` and `exchange` — the old `regular_cells` / `irregular_cells` split is gone, since the two
  now run inside one OpenMP region over the patches.
- The rest of the #710 checklist is untouched: **A4**, **A5**, the **M0** measurement, and all of phases B
  and C remain open.

### Alternative solutions

- **A4 as written in #710** — hold the MF alias shells as reusable members on `AmrMesh` — was implemented
  and then reverted. It saved one `LevelData` construction plus one `RefCountedPtr` counter per level per
  phase and nothing else: `LevelData::define` has no early-out for an unchanged layout, so every `aliasMF`
  call clears and rebuilds the exchange copier — a neighbour walk over every local box, allocating a
  `MotionItem` per pair — and re-runs `allocateGhostVector`. Two `malloc`s against O(boxes × neighbours)
  that are paid either way did not justify mutable state on `AmrMesh`. `CD_AmrMesh.H` is unchanged from
  `main`.
- **Widening the member buffer on demand and keeping it wide** was the first version of A3. See above for
  why it is a max rather than a current.
- **Deleting the per-level exchange outright** instead of guarding it with `a_doExchange` does not work:
  `AmrMesh::interpGhost(LevelData<EBCellFAB>&, …)` calls `interpolate` directly and relies on it. Moving
  the exchange into that caller was the other option; the flag keeps the "interpolate leaves valid ghost
  cells filled" contract as the default, so a new caller has to opt out deliberately.
- **Fusing the (L−1) `copyTo`s into one schedule** is the actual win in #710, but it is phase C. This PR
  deliberately stops at the point where the remaining round trips are one per level, which is the input
  that phase C's operation list is built from.

### PR-review checklist

Mandatory:

Optional:

### DIFF
--- a/Source/AmrMesh/CD_AmrMesh.H
+++ b/Source/AmrMesh/CD_AmrMesh.H
@@ -2399,8 +2399,10 @@ protected:
 
   /**
    * @brief Alias multifluid cell data into the reusable single-phase views m_mfAliasGas and m_mfAliasSolid.
-   * @details Only the phases that exist on the input realm are aliased; the view for a phase without an EB index
-   * space is left alone, and callers must not use it. The views are write handles rather than copies:
+   * @details This is the shell-reusing counterpart of allocatePointer followed by alias: it keeps the level
+   * shells alive between calls and only re-aliases them, then defers the aliasing itself to alias(). Only the
+   * phases that exist on the input realm are aliased; the view for a phase without an EB index space is left
+   * alone, and callers must not use it. The views are write handles rather than copies:
    * MfCellAliasFactory hands out the EBCellFAB inside each MFCellFAB of a_data, so everything a caller writes
    * through m_mfAliasGas or m_mfAliasSolid lands in a_data. This routine only reads a_data, but a_data is
    * modified through the views it returns, which is why it is taken by non-const reference.
--- a/Source/AmrMesh/CD_AmrMesh.H
+++ b/Source/AmrMesh/CD_AmrMesh.H
@@ -2400,9 +2400,12 @@ protected:
   /**
    * @brief Alias multifluid cell data into the reusable single-phase views m_mfAliasGas and m_mfAliasSolid.
    * @details Only the phases that exist on the input realm are aliased; the view for a phase without an EB index
-   * space is left alone, and callers must not use it.
+   * space is left alone, and callers must not use it. The views are write handles rather than copies:
+   * MfCellAliasFactory hands out the EBCellFAB inside each MFCellFAB of a_data, so everything a caller writes
+   * through m_mfAliasGas or m_mfAliasSolid lands in a_data. This routine only reads a_data, but a_data is
+   * modified through the views it returns, which is why it is taken by non-const reference.
    * @param[in]     a_realm Realm name. Must exist.
-   * @param[in,out] a_data  Multifluid data to alias.
+   * @param[in,out] a_data  Multifluid data to alias. Written to through the resulting views, not by this call.
    */
   void
   aliasMultifluidCellData(const std::string& a_realm, MFAMRCellData& a_data) const;
--- a/Source/AmrMesh/CD_AmrMesh.H
+++ b/Source/AmrMesh/CD_AmrMesh.H
@@ -2155,6 +2155,20 @@ protected:
   mutable std::map<std::string, RefCountedPtr<Realm>> m_realms;
 
   /**
+   * @brief Reusable single-phase alias of the gas-phase part of multifluid cell data.
+   * @details The MFAMRCellData routines operate by aliasing the multifluid data into single-phase views and
+   * calling the single-phase routine. The views are shells only -- they never own the data -- so they are held
+   * here and re-aliased in aliasMultifluidCellData rather than reallocated on every call.
+   */
+  mutable EBAMRCellData m_mfAliasGas;
+
+  /**
+   * @brief Reusable single-phase alias of the solid-phase part of multifluid cell data.
+   * @details See m_mfAliasGas.
+   */
+  mutable EBAMRCellData m_mfAliasSolid;
+
+  /**
    * @brief Implicit functions
    */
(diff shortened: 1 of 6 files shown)


rmrsk:
@claude review

rmrsk:
@claude review

rmrsk:
@claude review

PR #713, opened 2026-08-31
Transport the same diffusive flux in the Ito and CDR solvers

### LINKED ISSUE #706: Ito and CDR solvers transport different diffusive fluxes when D varies in space
## Summary

`ItoSolver`/`ItoKMC` and `CdrSolver` transport different diffusive fluxes when the diffusion coefficient
varies in space. The two agree only where `grad(D) = 0`, which makes particle and fluid results diverge in
exactly the regions where `D` changes fastest.

## Detail

`ItoKMCPhysics::isotropicDiffusion` (`Physics/ItoKMC/CD_ItoKMCPhysicsImplem.H:909`) draws an isotropic hop

```
hop = sqrt(2*D*dt) * N(0,1)
```

and `ItoKMCGodunovStepper::stepEulerMaruyamaParticles`
(`Physics/ItoKMC/TimeSteppers/ItoKMCGodunovStepper/CD_ItoKMCGodunovStepperImplem.H:1809`) applies

```
X^(k+1) = X^k + dt*v + hop
```

That is an Ito-interpretation Euler-Maruyama update. Its Fokker-Planck equation is

```
dn/dt = -div(v*n) + grad^2(D*n)          i.e. flux  v*n - grad(D*n)
```

whereas `CdrSolver` integrates

```
dn/dt = -div(v*n) + div(D*grad(n))       i.e. flux  v*n - D*grad(n)
```

The two differ by `div(n*grad(D))` and coincide exactly for constant `D`. Recovering the fluid form from an
Ito SDE requires the drift to be `v + grad(D)` (equivalently, integrating in the Hanggi-Klimontovich /
anti-Ito interpretation). There is no `grad(D)` term anywhere in `Source/ItoDiffusion` or `Physics/ItoKMC`.

## Why it matters

Nothing warns the user, and the discrepancy is largest where `D` varies fastest -- at a streamer head, where
`E/N` and hence `D(E/N)` change over a few cells. For atmospheric air the bias is on the order of
`|grad D| / |v| ~ 1-3%` at the head, signed such that the particle swarm lags the fluid solution. It is
invisible in any test with a constant diffusion coefficient, which is why it does not show up in the
convergence tests.

`Exec/Examples/ItoKMC/ComparisonCdrPlasma` compares the two models on the same problem and is affected
directly: with the tabulated `D(E/N)` the two models are integrating different equations, and freezing `D`
to a constant in both chemistry files is currently the only way to remove the ambiguity from that
comparison.

## Possible resolutions

1. Add the `grad(D)` drift correction to the Ito update so that the particle model transports
   `v*n - D*grad(n)`, matching `CdrSolver`. This needs `grad(D)` on the mesh and interpolated to particle
   positions, and would make the two models agree by construction.
2. Leave the Ito interpretation as it is and document it explicitly, in the `ItoSolver`/`ItoKMC` Sphinx
   pages and in the doxygen for `isotropicDiffusion`, so that users know the two solvers transport different
   fluxes and that comparisons should freeze `D`.

Which is correct is arguably a modelling question rather than a purely numerical one -- the density-gradient
expansion that defines `D` assumes a locally uniform field, where the two forms agree, so neither is
obviously right at a front that is not resolved by that assumption. But the two solvers in this code
currently make different choices without saying so, and the fluid convention (`D*grad(n)`) is the one the
rest of the code and the literature default to.

rmrsk:
# Summary

`ItoSolver`/`ItoKMC` and `CdrSolver` transported different diffusive fluxes wherever the diffusion coefficient
varied in space. This adds the `grad(D)` drift correction that makes the particle models transport
`v*n - D*grad(n)`, the same flux the fluid solvers integrate. 

Closes #706.

### Background

The particle update is an Ito-interpretation Euler-Maruyama step,

```
X^(k+1) = X^k + dt*v + sqrt(2*D*dt)*N(0,1),
```

whose Fokker-Planck equation is `dn/dt = -div(v*n) + grad^2(D*n)`, i.e. it transports the flux
`v*n - n*grad(D) - D*grad(n)`. `CdrSolver` integrates `dn/dt = -div(v*n) + div(D*grad(n))`, i.e. `v*n - D*grad(n)`.
The two differ by `div(n*grad(D))` and coincide exactly where `D` is constant in space, so nothing with a constant
diffusion coefficient can see the difference -- which is why it never showed up in the convergence tests. Recovering
the fluid form from an Ito SDE requires the drift to be `v + grad(D)` (equivalently, integrating in the
Hanggi-Klimontovich interpretation).

The discrepancy is largest where `D` varies fastest, i.e. at a streamer head, and is signed so that the particle
swarm lags the fluid solution.

### Solution

`ItoSolver` gains the mesh gradient and its interpolation; the time steppers apply it.

* **`ItoSolver`** -- `m_diffusionGradient`, a `SpaceDim` cell field allocated only when the solver is diffusive and
  the correction is enabled; `computeDiffusionGradient()`, which differentiates `m_diffusionFunction` through the
  existing `AmrMesh::computeGradient`/`EBGradient` path and then coarsens and fills ghosts;
  `interpolateDiffusionGradient()`, which puts `(grad D)(X_p)` on the `ItoParticle` scratch vector columns using the
  same interpolation the velocities already get. New option `ItoSolver.diffusion_grad_drift` (default `true`) and a
  `grad_dco` plot variable.
* **`ItoKMCGodunovStepper`** -- `diffuseParticlesEulerMaruyama` now computes the total explicit displacement,
  `hop + dt*grad(D)`, stores it on the particles, and places the `rho^dagger` particles with it.
* **`BrownianWalker`** -- the same term in its Euler step, applied under `isDiffusive()` rather than `isMobile()`,
  since it is a property of a varying `D` and not of having a velocity field.
* **`BrownianWalkerStepper::setDiffusion`** -- previously wrote `std::numeric_limits<Real>::max()` into the mesh
  diffusion field on the grounds that it was unused. It was not: the regression inputs ask for
  `ItoSolver.plt_vars = phi vel dco`, so every plot file those tests wrote carried `DBL_MAX` in the
  `diffusion_coefficient` component. It now holds `m_diffCo`.

**Why the correction goes with the hop rather than into the particle velocities.** `ItoKMCGodunovStepper` is
semi-implicit in the field, splitting the update as

```
X^(k+1) = X^k + [hop + dt*grad(D^k)]  +  dt*mu^k*E^(k+1)(X^k)
           explicit, known at time k       implicit
```

The semi-implicit Poisson operator `div((eps + dt*sigma^k) grad(phi^(k+1))) = -rho^dagger` is exact only because
`rho^dagger` is deposited from particles displaced by everything *except* the `dt*mu*E^(k+1)` term, and because
that remaining term is linear in `E^(k+1)`. `grad(D^k)` is lagged and field-independent, so folding it into the
explicit half leaves the Poisson operator, the conductivity, and `stepEulerMaruyamaParticles` untouched. This is
exactly what `computeDiffusionTermCDR` already does with the explicit `dt*div(D*grad(phi^k))` on the CDR side.

**Verification so far.**

* Full 2-D MPI `OPT=HIGH` build is clean.
* `Exec/Tests/BrownianWalker/DriftDiffusion` (`regression2d`, serial): comparing every value of every plot file
  against a pre-change run -- 16 932 240 values across 201 plot files and both levels -- 1 240 572 differ, and every
  single one goes from `1.7976931348623157e+308` to `0.01`, the configured `BrownianWalker.diffco`. Nothing else
  moves, so the trajectories are bit-identical across all 200 steps and 40 regrids. `D` is uniform there, so
  `grad(D)` is identically zero and the new path is provably neutral.
* `Exec/Tests/ItoKMC/JSON` (`regression2d`): runs to completion, no NaN, no assert. With the tabulated `D(E/N)`,
  `grad(D)` is finite and non-zero (`|grad D|` up to 1.4 m/s on that mesh).
* Sphinx builds clean under `-W --keep-going`; all fifteen `Ito.rst` `literalinclude` ranges were relocated by exact
  block match and re-verified after `clang-format`.

**Not yet verified.** `Exec/Examples/ItoKMC/ComparisonCdrPlasma` has not been rerun, so the size and sign of the
effect at a developed streamer head are not confirmed here -- the `ItoKMC/JSON` regression is too coarse and stops
too early to show it.

### Side-effects

* **Results change.** The correction is on by default, so every simulation whose diffusion coefficient varies in
  space -- which is every model using a tabulated `D(E/N)` -- moves. Results with a constant `D` are bit-identical.
* Plot files from the BrownianWalker tests previously contained `DBL_MAX` in the `dco` component and now contain
  the real coefficient.
* An `ItoKMCPhysics` diffusion model now sees `grad(D)` in `p.scratch_{x,y,z}` when the stepper calls it, instead
  of the previous step's stale hop. Documented on the `DiffusionFunction` alias.
* Every `.inputs` file that configures an `ItoSolver` gains `ItoSolver.diffusion_grad_drift`, since the solver
  reads it with `pp.get`. The key is deliberately spelled `diffusion_grad_drift` rather than
  `diffusion_gradient_drift` so that it fits inside the column width `format-input-files` aligns those blocks to;
  the longer name re-aligned every `ItoSolver` block in all 30 files, about 1400 lines of pure whitespace.
* The option is read once at setup, not at runtime -- it decides whether `m_diffusionGradient` is allocated at all.
  The Sphinx caption claiming every `ItoSolver` option is run-time configurable is corrected accordingly (`seed`
  was already an exception).
* **No new types or containers.** The mesh gradient is an `EBAMRCellData` following the existing
  `m_velocityFunction` allocate/`allocatePointer` pattern, and the per-particle storage reuses the existing
  `ItoParticle::scratch_x/y/z` columns rather than adding payload columns, so the per-particle footprint and the
  checkpoint format are unchanged. Those columns carry `grad(D)` only from the interpolation to a few lines later
  in the same loop iteration -- a strictly shorter lifetime than the hop they already held.
* Known limitations, documented rather than fixed: the correction is defined for a scalar isotropic `D` (a tensor
  would need `div(D)`); it is meaningless for a solver that assigns particle diffusion coefficients parametrically
  from particle energy, since there is then no mesh field whose gradient means anything (nothing in the tree does
  this -- `ItoSolver::updateDiffusion()` has no callers); and the `ItoSolver` time step estimates do not include it.
* `PLAN.md` at the repository root is the working plan for this change. Say the word and I will drop it from the
  branch.

### Alternative solutions 

**Put `grad(D)` into the particle velocity columns**, so that it is literally the drift velocity and
`computeAdvectiveDt` picks it up for free. Rejected on three counts: it would add a term that is not proportional
to `E^(k+1)` to the implicit half of the semi-implicit split, which is the half the Poisson operator was built for;
`interpolateVelocities` returns early for immobile solvers and `setItoVelocityFunctions` only fills the velocity
function for `Z != 0`, so diffusive-but-immobile neutral species would silently lose the term; and the stepper
would still need `grad(D)` separately for `rho^dagger`, so it would have to be interpolated twice and carefully not
double-counted. The only thing lost is the time step bound, and that is a couple of percent of a limit the
diffusive rule already dominates.

**Resolution 2 from #706 -- leave the Ito interpretation alone and document it.** Defensible as modelling: the
density-gradient expansion that defines `D` assumes a locally uniform field, where the two forms agree, so neither
is obviously right at an underresolved front. Not taken, because the two solvers in this code were making different
choices without saying so, and `D*grad(n)` is the convention the rest of the code and the literature default to.

**A limiter on `|dt*grad(D)|` near the EB**, where the field boundary layer can make `D(E/N)` vary sharply across
cut cells. Deliberately not added: the correction is a physical drift rather than a stability hazard, and a knob
should not go in ahead of evidence that it is needed.

### PR-review checklist

Mandatory:

Optional:


rmrsk:
@claude review

rmrsk:
@claude review

PR #716, opened 2026-08-31
Cut the unconditional AMR-wide work out of the ItoKMC reaction advance

### LINKED ISSUE #715: `ItoKMCStepper::advanceReactionNetwork` is dominated by unconditional halo deposition and redundant AMR synchronization
## Summary

Coarse `CD_Timer` instrumentation of `ItoKMCGodunovStepper::advance` on ~470 ranks shows that the reaction advance spends almost none of its time on chemistry. The KMC kernels are 8% of the call; 75% is `reconcileCdrDensities` and the photoionization deposit that precedes it — two blocks that perform AMR-wide, communication-bound work **unconditionally**, whether or not there is anything to do.

Measured split of `advanceReactionNetwork` (one time step, rank 0 local, global max in the tail):

| Event | Loc. (s) | Loc. (%) | Min (s) | Max (s) | Avg (s) |
|---|---|---|---|---|---|
| Compute PPC | 0.0037 | 0.63 | 0.0020 | 0.0040 | 0.0033 |
| Copy PPC back | 0.0001 | 0.02 | 0.0001 | 0.0016 | 0.0004 |
| KMC kernels | 0.0474 | 8.12 | 0.0114 | 0.1336 | 0.0538 |
| Reconcile particles | 0.0925 | 15.84 | 0.0045 | 0.3678 | 0.0891 |
| **Reconcile CDR** | **0.4399** | **75.38** | 0.1110 | 0.5633 | 0.4372 |

Two caveats on reading that table, both worth fixing separately:

- `Loc. (%)` is a fraction of **rank 0's** local total, not of the global.
- The `Dev.` column is not a standard deviation — `CD_TimerImplem.H` computes `sigma += sqrt(pow(x - avg, 2))` (i.e. `|x - avg|`) and then takes `sqrt(sigma/numRank)`, so its units are √s. Read `Dev²` as a mean absolute deviation.

Part of the 0.44 s is genuine load-imbalance absorption rather than work: `Reconcile CDR` is the first region after two badly imbalanced compute phases and it is full of collectives. The signature is visible in the ranks — rank 120 is the **max** in `Reconcile particles` (0.3678) and the **min** in `Reconcile CDR` (0.1110), and the per-rank sums flatten to ~0.48–0.53 across the board. Separating wait from work needs timed barriers between the events; that is not part of this issue, but it means the numbers above are an upper bound on the real work.

### A. The photoionization deposit runs for every CDR species even when there is nothing to deposit

`CD_ItoKMCStepperImplem.H:4038`:

```cpp
for (int i = 0; i < m_cdrPhotoiProducts.size(); i++) {
  m_cdrPhotoiProducts[i]->organizeParticlesByPatch();
  m_amr->depositWeight(m_particleScratch1, m_particleRealm, m_plasmaPhase,
                       *m_cdrPhotoiProducts[i], DepositionType::NGP,
                       CoarseFineDeposition::Halo, false);
  m_amr->copyData(m_fluidScratch1, m_particleScratch1);
  ...
}
```

There is no early-out. `m_cdrPhotoiProducts` is sized `numCdrSpecies` (`:616`) and is populated only by the model hook `m_physics->reconcilePhotoionization(...)` (`:4723`). Many plasma models route photoionization products to Ito species rather than CDR species, in which case these containers are empty on every step and the entire block is overhead.

`CoarseFineDeposition::Halo` lands in `depositHaloCore` (`Source/Particle/CD_EBAMRParticleMesh.H`), which per species does a hierarchy-wide `DataOps::setValue`, a mask-particle exchange (`copyMaskParticles`), a per-level `exchange(..., EBAddOp())`, `addFineGhostsToCoarse` per refined level, and a FiCo re-deposit plus `addFiCoDataToFine` per refined level — roughly `3L − 1` synchronizing stages per species, followed by a cross-realm `copyData` (MPI copier when `dual_grid = true`) and hierarchy-wide `volumeScale`/`incr`.

**Fix A.** Skip species whose product container is globally empty. The test has to be *global*, because the deposition is collective and all ranks must make the same decision; one `Allreduce` over the whole vector answers it for every species at once. `ParallelOps::sum(Vector<long long int>&)` already exists (`CD_ParallelOps.H:220`).

```cpp
Vector<long long int> numPhotoiProducts(m_cdrPhotoiProducts.size());
for (int i = 0; i < m_cdrPhotoiProducts.size(); i++) {
  numPhotoiProducts[i] = static_cast<long long int>(m_cdrPhotoiProducts[i]->getNumberOfValidParticlesLocal());
}

ParallelOps::sum(numPhotoiProducts);

for (int i = 0; i < m_cdrPhotoiProducts.size(); i++) {
  if (numPhotoiProducts[i] == 0) {
    continue;
  }
  ...existing body...
}
```

Exactly equivalent: an empty container deposits identically zero, so `copyData`/`volumeScale`/`incr` add zero. Skipping leaves stale data in `m_particleScratch1`/`m_fluidScratch1`, but nothing reads them — both are overwritten by their next user.

### B. Two of the four hierarchy sweeps in `coarsenCDRSolvers()` have no reader

`CD_ItoKMCStepperImplem.H:4871`, called unconditionally from `reconcileCdrDensities` (`:4790`). Per CDR solver it runs `conservativeAverage(phi)`, `conservativeAverage(src)`, `interpGhostPwl(phi)`, `interpGhostPwl(src)` — four hierarchy-wide sweeps, each with its own exchange.

Chasing the consumers on the reaction-advance path:

- **`interpGhostPwl(phi)`** — the next consumers of `phi` ghost cells are `computeDivF`/`computeDivD`, and both are documented as *"a_phi is non-const because ghost cells will be interpolated in this routine"* (`CD_CdrSolver.H:200`, `:217`). On top of that, `stepEulerMaruyamaCDR` re-does `conservativeAverage(phi)` + `interpGhostPwl(phi)` itself at the top of every solver iteration (`CD_ItoKMCGodunovStepperImplem.H:1874`). Everything in between (`computeCdrConductivity`, `computeReactiveCdrParticlesPerCell`, plot/checkpoint) reads valid cells only.
- **`interpGhostPwl(src)`** — `src` ghost cells are never read at all. The only consumers of `getSource()` are plotting and `ItoKMCBackgroundEvaluator`, both valid-cell.
- `conservativeAverage(phi)` should stay: `computeCdrConductivity` reads coarse valid cells next step, before the transport re-averages.
- `conservativeAverage(src)` matters only for output fidelity and could arguably move to the plot path.

Related: `coarsenCDRSolvers()` already runs at the end of `stepEulerMaruyamaCDR` (`CD_ItoKMCGodunovStepperImplem.H:1896`), so it executes at least **twice per time step**, and the transport step's copy is immediately followed on the next iteration by the same average-and-interpolate pair at `:1874`.

**Fix B.** Give the function a `const bool a_interpGhosts` parameter so the transport-step call keeps its current behaviour while `reconcileCdrDensities` passes `false`. Removes 2 of the 4 sweeps per CDR solver from the reaction advance.

### C. `ParticleSoA::clear()` invalidates the cell sort of an already-empty leaf, forcing an O(cells) zero-fill

In the `Reconcile particles` region, `CD_ItoKMCStepperImplem.H:4024`:

```cpp
for (int i = 0; i < m_cdrPhotoiProducts.size(); i++) {
  m_cdrPhotoiProducts[i]->clearParticles();
  m_cdrPhotoiProducts[i]->organizeParticlesByCell();
}
```

`ParticleSoA::clear()` (`CD_ParticleSoA.H:911`) unconditionally sets `m_sorted = false`, so `sortByCell`'s `isSortedAgainst` early-out always misses and every leaf reaches `m_cellStart.assign(nCells + 1, 0)` (`CD_ParticleSoAImplem.H:323`) — which sits *above* the `m_size == 0` return on the next line. The result is one offset zeroed per cell of every patch on every level, per CDR species, per time step, on containers that are empty by construction at that point. The cost is O(cells), not O(particles).

**Fix C.** An already-empty leaf's cell sort is still valid — its CSR offsets are all zero either way — so only a clear that actually discards particles need invalidate it:

```cpp
clear() noexcept
{
  if (m_size > 0) {
    m_sorted = false;
  }

  m_size = 0;
}
```

Safe because `isSortedAgainst` also compares box/dx/probLo, so a regrid still forces the re-sort.

### D. `AmrMesh::copyData` deep-copies the `Copier` on every cross-realm call

`CD_AmrMeshImplem.H:125` declares `Copier copier;` and copy-assigns the cached one out of the realm copier maps. Chombo's `Copier::operator=` (`Chombo-3.3/lib/src/BoxTools/Copier.cpp:112`) pool-allocates a fresh `MotionItem` for every entry in all three motion plans, so this is a per-level allocation loop on every cross-realm copy.

**Fix D.** Hold a `const Copier*` and dereference at the `copyTo`. Not a factor in the profile above (`Copy PPC back` is 0.02%), but it is free and sits on a very hot generic path used across the whole code base.

## Status

A, B and C are implemented locally on top of the coarse `CD_Timer` instrumentation and build clean in 3-D MPI (`OPT=HIGH`). `Exec/Tests/ItoKMC/JSON` (which has two `"solver" : "cdr"` species, so it does exercise the CDR path) runs 5 steps on 2 ranks without assertion failures or NaNs. D is not implemented.

Expected impact is entirely on MPI-synchronizing stages rather than FLOPs: A removes `numCdrSpecies × (3L − 1)` deposition stages plus `numCdrSpecies × L` copier exchanges per step whenever photoionization produces no CDR products; B removes 2 of 4 hierarchy sweeps per CDR solver; C removes an O(local cells) zero-fill per CDR species per step.

## Suggested follow-ups (not in scope here)

- Split `Reconcile CDR` three ways (`photoi deposit` / `cross-realm copy` / `coarsen CDR`) to attribute the remaining time between A and B.
- Add timed `ParallelOps::barrier()` events between the phases so imbalance wait is reported separately from work.
- Fix the `Dev.` column in `CD_TimerImplem.H`, and consider reporting `%` against the global rather than rank 0's local total.

rmrsk:
# Summary

`ItoKMCStepper::advanceReactionNetwork` spent almost none of its time on chemistry: the KMC kernels were ~8% of the call while `reconcileCdrDensities` and the photoionization deposit preceding it were ~75%, because both performed AMR-wide, communication-bound work unconditionally. This PR makes that work conditional on there being something to do, and stops three separate places from rediscovering information they already had.

Closes #715 

### Background

Coarse `CD_Timer` instrumentation of `ItoKMCGodunovStepper::advance` on ~470 ranks gave this split of one reaction advance (see #715 for the full table and its caveats):

| Event | Loc. (s) | Loc. (%) |
|---|---|---|
| Compute PPC | 0.0037 | 0.63 |
| Copy PPC back | 0.0001 | 0.02 |
| KMC kernels | 0.0474 | 8.12 |
| Reconcile particles | 0.0925 | 15.84 |
| **Reconcile CDR** | **0.4399** | **75.38** |

Part of the 0.44 s is load-imbalance absorption rather than work — `Reconcile CDR` is the first region after two badly imbalanced compute phases and is full of collectives — so the number is an upper bound. Separating wait from work needs timed barriers and is not attempted here.

### Solution

**Skip the photoionization deposit for CDR species with no products.** The photoionization products are *point particles*: `ItoKMCPhysics::reconcilePhotoionization` appends one per target species at the absorbed photon's position, and only afterwards are they deposited on the mesh and added into `m_fluidCdrPPC`. `m_cdrPhotoiProducts` is sized `numCdrSpecies` unconditionally, so a container exists for every CDR solver whether or not any photo-reaction can ever populate it. An empty container deposits identically zero, so the whole block — a `CoarseFineDeposition::Halo` deposit (roughly `3L-1` synchronizing stages per species), a cross-realm `copyData`, and hierarchy-wide `volumeScale`/`incr` — adds nothing. The test is a single `ParallelOps::sum` over the whole species vector, because the work being skipped is collective and every rank must reach the same verdict. It is a per-species, per-step runtime question, not an inspection of the reaction set: it covers both a species no photo-reaction targets and a step in which no photon was absorbed into that channel.

**Make the ghost-cell interpolation in `coarsenCDRSolvers()` optional.** The function is called at least twice per time step and ran four hierarchy sweeps per CDR solver. Two of them have no reader on the reaction path: the next consumers of the density ghost cells are `CdrSolver::computeDivF`/`computeDivD`, both documented as interpolating ghost cells themselves, and the source-term ghost cells are never read at all (`getSource()`'s consumers — plotting and `ItoKMCBackgroundEvaluator` — are valid-cell only, and the plot path fills its own). Both `conservativeAverage` calls are kept: `computeCdrConductivity` reads coarse valid cells before anything re-averages them. The transport-step call keeps its current behaviour.

**Stop rediscovering cell sorts that are already known.** Three linked changes:

- `ParticleSoA::clear()` no longer invalidates the sort when there was nothing to drop. An already-empty leaf's CSR is all-zero regardless of the cells it was built against, so `sortByCell()` was falling through its early-out and rebuilding `nCells+1` offsets per patch, per level, per species, per step on containers that are empty by construction.
- `rebuildLeafFromCells` writes the per-cell scratch back in Fortran cell order, so the rebuilt leaf *is* cell-sorted; it now prefix-sums the per-cell counts it already walks and adopts the mapping instead of swapping in an unsorted container.
- `ItoSolver::extractIntoMergeContainer` copies particles in index order, so the reduced `ItoMergeParticle` leaf has the same CSR as its `ItoParticle` source. Carrying it across lets `applyCellMerger`'s `organizeParticlesByCell()` take its early-out, removing a full counting sort plus column permute per species, per patch, per merge step. All four cell-based merge methods route through `applyCellMerger`.

**Stop deep-copying the `Copier` in `AmrMesh::copyData`.** `Copier::operator=` pool-allocates a fresh `MotionItem` for every entry in all three motion plans, so the cached copier was rebuilt on every cross-realm copy. Holding a `const Copier*` and dereferencing at the `copyTo` is free and sits on a hot generic path used across the code base.

### Side-effects

- **`ItoKMCStepper::coarsenCDRSolvers()` changes signature** to `coarsenCDRSolvers(const bool a_interpGhosts)`. It is `virtual`; there are no in-tree overrides, but this is a compile-visible break for any downstream subclass that overrides it.
- **`ParticleSoA` gains two methods**, `adoptCellSort()` (3 uses) and `cellOffsets()` (1 use). No new types or containers are introduced: the early-out uses the existing `Vector<long long int>` with the existing `ParallelOps::sum` overload, and `adoptCellSort()` writes the CSR fields `sortByCell()` already maintains rather than adding a parallel representation. `adoptCellSort()` asserts the adopted mapping against `checkCellSort()`, the same debug-mode verifier that guards `sortByCell()`'s own early-out, so a caller that lays particles out inconsistently fails loudly in a `DEBUG` build instead of silently mis-binning them.
- Skipping the deposit leaves stale data in `m_particleScratch1`/`m_fluidScratch1`. Nothing reads them — both are fully overwritten by their next user.
- The expected gain is in MPI-synchronizing stages rather than FLOPs, and the photoionization early-out only pays off for models that route photoionization to Ito species (or to a subset of the CDR species). No performance measurement at scale is included in this PR; the numbers above are the pre-change profile only.

### Alternative solutions 

- **Precompute which CDR species can ever receive photoionization products**, from the reaction set at define time, instead of testing the containers each step. Rejected: it would miss the transient case (a species that can receive products but received none this step), it would need invalidating whenever the physics is redefined, and the runtime test costs one `Allreduce` of a vector already sized by the species count.
- **Drop the two `interpGhostPwl` calls from `coarsenCDRSolvers()` outright** rather than making them conditional. Rejected: the transport-step caller is not covered by the same reasoning, and a flag keeps that path bit-for-bit unchanged.
- **Move `conservativeAverage(src)` to the plot path**, since the source term matters only for output fidelity. Not done here — it is a behavioural question about what a checkpoint should contain, not a redundancy.
- **Make `sortByCell()` cheap for empty containers** instead of preserving the sorted-empty state through `clear()`. Rejected: the all-zero CSR still has to be materialized for the right cell count, so the O(cells) fill remains; keeping the state is what actually removes it.
- **Leave `rebuildLeafFromCells` unsorted and only fix `clear()`**, as originally sketched in #715. This does not work: the rebuild swaps in a container whose sort flag is already false and whose offsets have been swapped away entirely, so the flag is false by the time `clear()` sees it and the next sort reallocates rather than merely refilling.

### PR-review checklist

Mandatory:

Optional:

---

**Verification performed while preparing this PR** (recorded here for the reviewer; the checklist above is left for the author):

- `Exec/Tests/ItoKMC/JSON`, 20 steps, clean-built both sides, at **2 and 4 ranks**: `h5diff -c` reports zero differences on all ten plot files. The change is exactly neutral in floating-point terms — an empty deposit adds identically zero and the dropped sweeps only touch unread ghost cells — so dataset-identical output is the right bar rather than a tolerance. This test exercises both branches of the early-out: its only photoionization is `Y + (O2) -> e + O2+`, so the CDR species `O2+` deposits while the CDR species `N2+` is skipped every step.
- Built with `DEBUG=TRUE OPT=TRUE`, so `CH_assert` is live (it is compiled out under `OPT=HIGH`): `adoptCellSort()`'s `checkCellSort()` assertion held for every rebuild site and for the merge-container adopt.
- 3-D MPI `OPT=HIGH` builds clean and runs 5 steps on 4 ranks with no NaNs.
- `pre-commit` passes clang-format, reuse, codespell, cppcheck and the doxygen hook. No `.. literalinclude::` directive references any of the touched files, so the shifted line numbers are inert.

### DIFF
--- a/Physics/ItoKMC/CD_ItoKMCStepper.H
+++ b/Physics/ItoKMC/CD_ItoKMCStepper.H
@@ -1601,9 +1601,16 @@ protected:
 
   /**
    * @brief Coarsen data for CDR solvers
+   * @details Always averages the densities and source terms down the hierarchy -- the coarse valid cells
+   * are read before anything re-averages them. The ghost-cell interpolation is optional because it has no
+   * reader on the reaction-advance path: the next consumers of the density ghost cells are
+   * CdrSolver::computeDivF/computeDivD, both of which interpolate the ghost cells themselves, and the
+   * source-term ghost cells are never read at all (getSource()'s consumers -- plotting and
+   * ItoKMCBackgroundEvaluator -- are valid-cell only, and the plot path fills its own ghost cells).
+   * @param[in] a_interpGhosts Whether to interpolate the ghost cells after averaging.
    */
   virtual void
-  coarsenCDRSolvers() noexcept;
+  coarsenCDRSolvers(const bool a_interpGhosts) noexcept;
 
   /**
    * @brief Resolve particle injection at EBs.
--- a/Physics/ItoKMC/CD_ItoKMCStepperImplem.H
+++ b/Physics/ItoKMC/CD_ItoKMCStepperImplem.H
@@ -61,24 +61,35 @@ binLeafToCells(std::vector<ParticleSoA<P, Traits>>& a_cells,
 }
 
 // Rebuild a_leaf from the per-cell scratch vector (concatenate all cells, then swap into the leaf).
+// The cells are written back in Fortran cell order, so the rebuilt leaf is cell-sorted by construction
+// and adopts the CSR mapping rather than letting the next sortByCell() rediscover it.
 template <typename P, typename Traits>
 inline void
-rebuildLeafFromCells(ParticleSoA<P, Traits>& a_leaf, const std::vector<ParticleSoA<P, Traits>>& a_cells) noexcept
+rebuildLeafFromCells(ParticleSoA<P, Traits>&                    a_leaf,
+                     const std::vector<ParticleSoA<P, Traits>>& a_cells,
+                     const Box&                                 a_box,
+                     const Real                                 a_dx,
+                     const RealVect&                            a_probLo) noexcept
 {
-  ParticleSoA<P, Traits> rebuilt;
+  CH_assert(a_cells.size() == static_cast<std::size_t>(a_box.numPts()));
 
-  // Pre-reserve the known total so the append loop allocates once instead of growing
-  // geometrically (O(log N) reallocations) on every reaction-kernel step.
-  std::size_t total = 0;
-  for (const ParticleSoA<P, Traits>& cell : a_cells) {
-    total += cell.size();
+  // Prefix-sum the per-cell counts into the CSR start offsets. This also gives the known total, so the
+  // append loop below allocates once instead of growing geometrically (O(log N) reallocations) on every
+  // reaction-kernel step.
+  std::vector<std::size_t> cellStart(a_cells.size() + 1, 0);
+  for (std::size_t c = 0; c < a_cells.size(); c++) {
+    cellStart[c + 1] = cellStart[c] + a_cells[c].size();
   }
-  rebuilt.reserve(total);
+
+  ParticleSoA<P, Traits> rebuilt;
+  rebuilt.reserve(cellStart.back());
 
   for (const ParticleSoA<P, Traits>& cell : a_cells) {
     rebuilt.append(cell);
   }
+
   a_leaf.swap(rebuilt);
+  a_leaf.adoptCellSort(a_box, a_dx * RealVect::Unit, a_probLo, std::move(cellStart));
 }
 
 } // namespace
@@ -4035,7 +4046,30 @@ ItoKMCStepper<I, C, R, F>::advanceReactionNetwork(const EBAMRCellData& a_electri
   // the photoionization products are created in reconcileParticles and  deposited them on the particle realm, but
   // the resulting products need to end up on in the correct data component in m_fluidCdrPPC.
   CH_START(t5);
+
+  // Species whose product container is empty on EVERY rank deposit identically zero, so the whole block
+  // below -- a halo deposition (which alone is ~3L-1 synchronizing stages), a cross-realm copy, and two
+  // hierarchy-wide passes -- adds nothing to m_fluidCdrPPC. Many plasma models route photoionization to
+  // Ito species rather than CDR species, in which case that is every species on every time step.
+  //
+  // The test must be global because the work being skipped is collective: every rank has to reach the
+  // same verdict for every species. One Allreduce over the whole vector answers it for all of them, and
+  // the local counts it reduces are free (the containers track their own size). Skipping leaves stale
+  // data in m_particleScratch1/m_fluidScratch1, which is harmless -- both are fully overwritten by their
+  // next user.
+  Vector<long long int> numPhotoiProducts(m_cdrPhotoiProducts.size(), 0LL);
+
   for (int i = 0; i < m_cdrPhotoiProducts.size(); i++) {
+    numPhotoiProducts[i] = static_cast<long long int>(m_cdrPhotoiProducts[i]->getNumberOfValidParticlesLocal());
+  }
+
+  ParallelOps::sum(numPhotoiProducts);
+
+  for (int i = 0; i < m_cdrPhotoiProducts.size(); i++) {
+    if (numPhotoiProducts[i] == 0LL) {
+      continue;
+    }
+
     m_cdrPhotoiProducts[i]->organizeParticlesByPatch();
 
     m_amr->depositWeight(m_particleScratch1,
@@ -4640,7 +4674,7 @@ ItoKMCStepper<I, C, R, F>::reconcileParticles(const EBCellFAB& a_newParticlesPer
 
     ParticleContainer<ItoParticle>& solverParticles = solverIt()->getParticles(ItoSolver::WhichContainer::Bulk);
 
-    rebuildLeafFromCells(solverParticles[a_level][a_din], itoCells[idx]);
+    rebuildLeafFromCells(solverParticles[a_level][a_din], itoCells[idx], a_box, a_dx, probLo);
   }
 
   for (auto solverIt = m_rte->iterator(); solverIt.ok(); ++solverIt) {
@@ -4648,14 +4682,14 @@ ItoKMCStepper<I, C, R, F>::reconcileParticles(const EBCellFAB& a_newParticlesPer
 
     ParticleContainer<Photon>& solverSourcePhotons = solverIt()->getSourcePhotons();
 
-    rebuildLeafFromCells(solverSourcePhotons[a_level][a_din], sourcePhotonCells[idx]);
+    rebuildLeafFromCells(solverSourcePhotons[a_level][a_din], sourcePhotonCells[idx], a_box, a_dx, probLo);
   }
 
   // Rebuild the CDR photoionization-product leaves from the mutated per-cell scratch.
   for (auto solverIt = m_cdr->iterator(); solverIt.ok(); ++solverIt) {
     const int idx = solverIt.index();
 
-    rebuildLeafFromCells((*m_cdrPhotoiProducts[idx])[a_level][a_din], cdrCells[idx]);
+    rebuildLeafFromCells((*m_cdrPhotoiProducts[idx])[a_level][a_din], cdrCells[idx], a_box, a_dx, probLo);
   }
 }
 
@@ -4787,7 +4821,9 @@ ItoKMCStepper<I, C, R, F>::reconcileCdrDensities(const EBAMRCellData& a_newParti
       }
     }
 
-    this->coarsenCDRSolvers();
+    // No ghost interpolation: nothing reads the density or source-term ghost cells between here and the
+    // next transport step, which fills them itself. See coarsenCDRSolvers().
+    this->coarsenCDRSolvers(false);
   }
 }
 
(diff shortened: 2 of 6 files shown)


rmrsk:
@claude review

PR #717, opened 2026-08-31
Update Chombo

rmrsk:
Phase B2 of #710. Independent of the #25/#26/#27 bump in #712 — it neither needs those changes nor is needed by them.

# Summary

### Background

`BoxLayoutData<T>::unpackReceivesToMe` did one `MPI_Waitall` over every posted receive and only then unpacked anything:

```cpp
MPI_Waitall(numReceives, &m_receiveRequests[0], &m_receiveStatus[0]);
for (unsigned int i=0; i<m_toMe.size(); ++i) { ... a_op.linearIn(...); }
```

The rank therefore sits inside the wait while data that has already landed goes untouched, and only starts unpacking once the *slowest* peer has arrived. The exposed cost is `slowest peer + entire unpack` where it could be `slowest peer` alone, with the unpack of every other peer hidden under the tail. This is on every `exchange()` and every `copyTo()` in the code base.

The mechanism needed to do better already existed. `postReceivesToMe` coalesces `m_toMe` by `procID` into one message per peer, so the arrival of a single message already corresponds to a known, contiguous run of `m_toMe` entries — that correspondence simply was not recorded anywhere.

### Solution

Bumps `Submodules/Chombo-3.3` from `f87edbf` to `00f7268`, picking up one PR against `chombo-discharge/Chombo-3.3`:

| Chombo PR | Change | Reach |
|---|---|---|
| [#28](https://github.com/chombo-discharge/Chombo-3.3/pull/28) | `unpackReceivesToMe` drives the receives with `MPI_Waitsome` and unpacks each peer's data as its message lands | every `copyTo` and every `exchange` |

`postReceivesToMe` now records, per peer, which contiguous run of `m_toMe` entries that peer's message feeds and which requests carry it. `unpackReceivesToMe` loops on `MPI_Waitsome`, decrements a per-group pending count for each completed request, and unpacks a group once its count reaches zero.

Three details from the upstream PR that are worth repeating here:

- **The pending count is not decoration.** A message larger than `CH_MAX_MPI_MESSAGE_SIZE` is split across several requests into one contiguous buffer slice, so unpacking on the first completion would read a partially-arrived buffer. A group is unpacked only when every request carrying it has completed.
- **Oversized-message chunks are pushed inline** instead of being collected in a `std::list` and appended after every other request, which used to scatter a group's requests across the array. The `MPI_Irecv` call order is unchanged, and matching does not depend on it in any case: groups differ by source rank, chunks within a group by tag.
- **The zero-size entries left behind by coalescing stay inside their group's range.** They carry valid `bufPtr`s from `allocateBuffers` and must still be unpacked; only their `size` was folded into the run's first entry.

`unpackReceivesToMe_append`, the particle path, still does its own `MPI_Waitall`. It appends to a `LayoutData` and is explicitly not thread-safe, so it was left alone.

No chombo-discharge source changes: the submodule pointer is the whole diff.

### Side-effects

- **Behaviour changes everywhere, in Chombo, not here.** Nothing in `Source/`, `Physics/` or `Geometries/` is touched, but every ghost exchange and every `copyTo` in the code base goes through the rewritten routine.
- **Results are expected to be bit-identical, with no tolerance to fall back on.** The change reorders *when* buffers are unpacked, not what is computed from them, and it reassociates no floating-point operation. Any `h5diff` hit would be a real defect — a mis-grouped unpack writing the wrong bytes into ghost cells — rather than numerical noise.
- **The failure mode is quiet.** A mis-grouped unpack does not crash; it silently corrupts ghost cells and the run continues. That is why bit-identity across the full suite, rather than "it ran to completion", is the check.
- **The chunked (`> CH_MAX_MPI_MESSAGE_SIZE`) path remains unexercised.** No test in the suite comes near a 30 MB single message, so the pending-count logic for split messages is covered by inspection only. That branch is the one worth a targeted test.
- `CopierBuffer::m_toMeUnpack` — declared for what looks like exactly this purpose and referenced nowhere in the tree — is replaced by the two members the scheme actually needs (`m_recvGroups`, `m_recvGroupOfRequest`). Net member count +1.
- **No performance measurement.** #710's M0 was dropped as a local task because the timing needs to run at scale (10k+ cores). The gain is bounded by unpack cost relative to wait cost, which is exactly what M0 would have measured, so the size of the win is currently unknown. Nothing here should be merged on a performance claim.

**Verification.** The same protocol as #712: benchmarks generated from a clean build of the pre-bump tree (`82383f84` + Chombo `f87edbf`), then a clean build of this bump compared against them with `h5diff` in 2-D and 3-D. All outputs are bit-identical.

Two traps worth repeating for anyone reproducing this:

- `make pristine` between the two builds, never an incremental one. Chombo headers change across the bump, and an incremental rebuild leaves stale objects whose vtable mismatch produces segfaults that look exactly like logic bugs.
- `-cores` sets both `make -j` and `mpirun -np`, and the rank count feeds the domain decomposition, so the benchmark run and the comparison run must use the same value.

### Alternative solutions

- **Leave the `MPI_Waitall` in place and reduce the number of exchanges instead.** That is what the rest of #710 is about, and it is complementary rather than an alternative: this change lowers the cost of an exchange that still has to happen.
- **Unpack a group on the first completed request rather than tracking a pending count.** Simpler, and correct for every message the current test suite produces, but wrong the moment a message exceeds `CH_MAX_MPI_MESSAGE_SIZE` — and wrong silently. Rejected.
- **Keep the oversized chunks in the side list and record non-contiguous request ranges per group.** Would preserve the old request layout exactly, at the price of a per-group index list instead of a `(begin, count)` pair. Rejected: the request array order is not observable — matching goes by source rank and tag — so there is nothing to preserve.
- **Copy each group's `bufEntry`s into per-group vectors** (what the unused `m_toMeUnpack` member appears to have been for). Rejected in favour of ranges into `m_toMe`, which already holds the entries in exactly the group order.

### PR-review checklist

Mandatory:

Optional:


rmrsk:
@claude review

PR #718, opened 2026-09-02
Reflect diffusion hop near EBs to avoid spurious sheaths

rmrsk:
# Summary

Gives the Ito diffusion hop a boundary condition at the EB, fixes a non-terminating ray-cast intersection, and makes `mirror` the cut-cell deposition in every `ItoSolver` configuration.

### Background

`ItoKMCGodunovStepper::diffuseParticlesEulerMaruyama` displaces particles by `sqrt(2*D*dt)*W` without asking whether the result lies inside the solid. A particle in the fluid can hop into the solid; none inside can hop back. The lost electrons deposit into covered cells, which carry no VoF and contribute nothing to `rho`, so the near-wall region drains every step. The result is a spurious positive space charge confined to the cut-cell layer, one cell wide at every resolution, present under every cut-cell deposition scheme and absent when the electron diffusion coefficient is zero.

Separately, `ItoSolver::intersectParticles` passed `tolerance = 0.0` into `AmrMesh::intersectParticlesRaycastIF`. The ray march declares a hit on `d < tolerance` with `d = |f(x)| >= 0`, so the branch can never be taken; on a path that does hit the surface the march converges geometrically and the loop does not terminate.

### Solution

**`ItoKMCGodunovStepper.diffusive_deposit = inside | cancel | reflect_rho | reflect_ito`**, defaulting to `reflect_ito`.

| | trajectory | rho^dagger deposit |
|---|---|---|
| `inside` | untouched | deposits into the covered cell; current behaviour |
| `cancel` | untouched | at the pre-hop position |
| `reflect_rho` | untouched | mirrored about the boundary |
| `reflect_ito` | mirrored | follows the corrected hop |

`reflectDiffusionHop` locates the crossing with the owning solver's own intersection routine (`intersection_alg`, `bisect_step`) so that reflection and the absorption test later in the advance agree on where the surface is, takes the normal from the gradient of the implicit function there, and mirrors the overshoot. It always brackets from the pre-hop position, which is known to lie in the fluid. One retry covers concave corners and thin features; a reflection that cannot reach the fluid falls back to the pre-hop position. Counters at `verbosity > 5`.

`reflect_ito` is the default because it is closest to the observed behaviour. It makes the wall reflecting for the diffusive part of the step while it stays absorbing for the drift part, which is a split along the operator splitting rather than along anything physical. This is a deliberate approximation, documented as such on the enum: the physical anode sheath here is `D_e/v_drift ~ 0.5-1.3 um` against `dx >= 1.2 um` at maximum refinement, and the ions are configured immobile and non-diffusive, so the structure is not representable in this model at any resolution. What `inside` produces in its place is `sqrt(2*D*dt)` wide, 2-10x wider than the physical layer and tied to `dt` and `dx`.

**`ItoKMCGodunovStepper.rho_dagger_hop`** (default `true`) allows the stochastic hop to be left out of `rho^dagger` while the particle still takes it. The lagged `grad(D)` drift is never left out.

**Ray-cast tolerance.** `AmrMesh::intersectParticlesRaycastIF`'s `a_tolerance` is now a fraction of the level `dx`, scaled inside the level loop, and `ItoSolver` passes `1.E-3` — the same effective tolerance McPhoto already uses. `ItoSolver` is that function's only caller. New `ItoSolver::getBisectionStep()` beside the existing `getIntersectionAlgorithm()`.

**`ItoSolver.deposition_irreg = mirror`** in all 28 inputs files that overrode it, matching `CD_ItoSolver.options`.

**`Driver::coarsenTags`** restores `tags[l] >= coarsen(tags[l+1])` before tags reach `AmrMesh`. `TiledMeshRefine` injects proper nesting unconditionally, so one tag on level `l` can hold every level beneath it with no tags in between; when that tag is coarsened away the region collapses past the cell tagger's `max_coarsen_lvl`.

### Side-effects

- Results change wherever Ito particles diffuse near an EB. Benchmarks move.
- `Driver::coarsenTags` changes the grids for every cell tagger, so 2d and 3d benchmarks move independently of the ItoKMC work.
- The 22 `BrownianWalker` regressions move off `redistribute`. No checked-in configuration now exercises `native`, `ngp`, `redistribute` or `redistribute_blended`.
- `diffusive_deposit` and `rho_dagger_hop` are mandatory keys, added to the options file and the seven inputs that configure this stepper.
- `AmrMesh::intersectParticlesRaycastIF`'s `a_tolerance` changes meaning from an absolute length to a fraction of `dx`.

### Alternative solutions

- **Bound the hop length in cells.** Changes the diffusion physics everywhere rather than only where geometry demands it, and does not fix the one-way leak, only makes it rarer.
- **Absorb at the wall and route the charge to `sigma`.** Dead on an electrode: the BC enters through the Dirichlet condition and the surface charge never appears there.
- **First-passage (Brownian bridge) absorption**, `p = exp(-d0*d1/(D*dt))`, applied to the whole step at the intersection stage. This is the correct absorbing wall — the endpoint test is `O(sqrt(dt))`, per Peters & Barenbrug, Phys. Rev. E 66, 056701 (2002) — but it would faithfully reproduce a sheath this model cannot represent. Not pursued.

Known gaps, both untested: a hop passing clean through geometry thinner than `bisect_step`, and a reflected position that leaves the domain.

### PR-review checklist

Mandatory:

Optional:


rmrsk:
@claude review

PR #720, opened 2026-09-05
Consolidate the super-particle merge selectors onto scope and specifier axes

### LINKED ISSUE #721: Reorganize particle mergers
We have many of these, but these should probably be reorganized somewhat. For example, kd trees inside cells can partition on number, or weight, with placement either at centroid or sampled, or randomized.

It'd be better to reorganize this...

rmrsk:
# Summary

Super-particle merging in `ItoSolver` was selected by a single `merge_algorithm` string with eleven
values. That string carried three orthogonal decisions at once, so it had eleven names for what are
really four algorithms, and the specifiers that went with them (`kd_cell_*`, `nn_pair_*`,
`kd_split_weight_leaf_dx`) each applied to a different, unstated subset. This PR splits those
decisions onto named axes, removes one merge method that lost on every measured axis, and adds the
sub-cell placement work that motivated the reorganization in the first place.

Closes #721.

### Background

A merge study on a 512²/64³ uniform population, merged repeatedly, established two things.

The first is that a merge which places each group at its weighted centroid drives the particles onto
a sub-cell lattice. The kd tree bisects a cell into `T` near-equal sub-boxes and each leaf collapses
to its centre, so the surviving particles land on a `T^(1/D)` lattice that is *identical in every
cell* and therefore coherent across the whole domain, sharpening with every merge. The prediction
`k = T^(1/D)` was confirmed across five configurations and against patch-size commensurability.
Placing the merged particle at a weight-proportional sample of the group's own members is
distribution-preserving and removes the lattice entirely: after 100 merges in 2-D, sub-cell
anisotropy 1.8%, `E[(u-1/2)²] = 0.0833` against the uniform 1/12, `N_eff` 16.00 of 16.

The second is that cell-scale density error comes *only* from mass moving across cell faces — every
per-cell method reproduces `sigma_post/sigma_pre = 1.000` exactly.

`equal_weight_kd`, `equal_weight_kd_sampled` and `reinitialize_bvh` were the first cut at exposing
that: one partition rule, three placements, spelled as three selectors. Issue #721 asked for those
to become one selector plus options, and doing that made the same problem visible one level up — the
whole selector list was axes flattened into strings.

### Solution

**One selector for the scope, specifiers for the rest.** `merge_algorithm` becomes `merge_method`
and names only *what the merge groups particles over*: `kd_cell`, `kd_patch`, `kd_amr`, `nn_amr`,
`reinitialize`, `external`, `none`. Everything else moves to a specifier:

| was | is |
|---|---|
| `equal_weight_kd` / `_sampled` / `reinitialize_bvh` | `kd_cell` + `kd_placement = centroid`/`sample`/`random` |
| `kd_carve` | `kd_amr` + `kd_amr_boundary = carve` |
| `kd_skin_nn` | `kd_amr` + `kd_amr_boundary = nn` |
| `nn_pair_tree` / `_hash` / `_onecell` | `nn_amr` + `nn_search = tree`/`hash`/`onecell` |
| `kd_cell_partition` / `kd_cell_placement` | `kd_partition` / `kd_placement`, for **every** kd method |
| `nn_pair_iterate` / `_fallback` / `_max_cell_dist` / `_max_rounds` | `nn_iterate` / `nn_fallback` / `nn_max_cell_dist` / `nn_max_rounds` |
| `kd_split_weight_leaf_dx` | `kd_partition = hybrid` + `kd_hybrid_leaf_dx` |

Eleven selectors become seven. No selector is silently remapped: every legacy string aborts with its
replacement spelled out, the way the `equal_weight_kd` family already did.

**`kd_partition` now means one thing everywhere.** The patch and AMR builds already split at the
count median above a length scale and the weight median below it; the per-cell build had a hard
two-way switch. Both now take a single crossover length — `weight` is a crossover above any node's
size, `count` one below every node's size, `hybrid` is the crossover you name in
`kd_hybrid_leaf_dx`. One knob, three tree builds. `kd_cell` results are unchanged at those limits:
both the split rule and the splittability test reduce to exactly what they were.

**Splitting places its pieces where you tell it to.** The weight median balances a node by *dividing*
the median particle and emitting the daughters co-located; `kd_placement = sample` then inherits those
positions verbatim, so a duplicate propagates into further leaves and compounds. After 500 merges a cell
holding 16 particles held a median of **six distinct positions**. `N_eff` reads 16.00 throughout -- the
weights really are uniform -- so no existing metric reported it. `ItoSolver.split_placement` says where
the pieces go: `center` (today's behaviour and the default), `jitter` (drawn from the local mean
interparticle spacing `dx/ppc^(1/D)` around the parent) or `cell` (uniform in the owning cell). Every
piece stays inside the parent's own cell, so the merge still moves no mass across a face, and all three
fall back to `center` in a cut cell. The jitter kernel is *truncated* to the cell rather than clamped or
reflected -- reflection folds the overhanging kernel back onto the part that stayed, giving a corner
`2^D` times the density it should.

**`kd_partition = weight_capped`.** The whole-patch build may not divide a particle mid-tree, so one
heavier than a leaf's target weight sets a floor no partition can get under -- the reason `kd_patch`/
`kd_amr` sit at `N_eff` ~12 where `kd_cell` reaches 16. A pre-pass divides such particles before the tree
is built; it is local and needs no id arbitration, so it works for `kd_amr` too. Only crowded cells are
capped, and ghosts never are.

**`kd_placement` applies to `kd_patch` and `kd_amr`.** They previously always placed a leaf's
particle at its weighted centroid, which is precisely the lattice-forming choice the study
identified. All three emission sites (carve interior tier, carve boundary tier, `mergeKDInterior`)
now go through one `detail::kdPlaceMerged` helper.

**`nn_sfc` is removed** rather than renamed. It was the one cell-based method that neither equalized
weights nor preserved the sub-cell distribution, and it lost to the per-cell kd merge on every axis
measured. `ParticleManagement::makeSfcNearestNeighborMerger`, `detail::mergeAdjacentNearest` and
`ItoSolver::mergeNnSfc` go with it.

**`BrownianWalkerStepper` reads its options under a configurable prefix.** It was the only stepper in
the tree hardcoding its ParmParse prefix (`ItoSolver` uses `m_className`, `CellTagger` and
`ItoKMCStepper` use `m_name`), which forced a derived stepper to split its input script across two
prefixes. It and `BrownianWalkerSpecies` now take the prefix as a constructor argument defaulting to
`BrownianWalker`, so every existing input file is untouched.

**New example**, `Exec/Examples/BrownianWalker/ParticleMerge`. A two-level grid with a manual
refinement slab, 16 ppc with 32 fresh particles per cell seeded every step followed by a merge, and a
per-level report of the quantities the study used: sub-cell Fourier amplitudes, `E[(u-1/2)²]`,
per-cell density, block-phase response, `N_eff`, and achieved ppc. It is a `BrownianWalkerStepper`
application, so it runs real time steps through transport, remap, deposit, regrid and merge.

### Side-effects

- **Every input script in the tree changes.** All 34 files carrying an `ItoSolver` block are
  rewritten to the order in `CD_ItoSolver.options` — what to merge to, what merges it, then the `kd_*`
  and `nn_*` specifiers — with each file's own values carried across. Four keys that no longer read
  anything anywhere (`seed`, `max_diffusion_hop`, `halo_deposition`, `redistribute`) are dropped.
- **Two new enums**, `KDAmrBoundary` and `NNSearch` (1 use each in `ItoSolver`, plus their
  `*FromString` helpers), against four removed `ParticleMergeMethod` values. `NNSearch` also replaces
  the private `ItoSolver::NnPairSearchBackend`, so the net change is +2 types, -1 type, -4 enum values.
- **The merge regression tests are renamed** after the `(method, specifier)` pair each covers:
  `EqualWeightKD`→`KDCellCentroid`, `KDCellSampled`→`KDCellSample`, `ReinitializeBVH`→`KDCellRandom`,
  `KDCarve`→`KDAmrCarve`, `KDSkinNn`→`KDAmrNn`, `NNMerge{Tree,Hash,OneCell}`→`NNAmr{Tree,Hash,OneCell}`.
  Benchmark files are not checked in, so this costs a benchmark regeneration and nothing else.
  `SFCMerge` is deleted with `nn_sfc`; `KDAmrSample` is added to cover `kd_amr` with a non-centroid
  placement.
- **`kd_placement` on `kd_patch`/`kd_amr` can move a particle across a cell face.** Those leaves are
  deliberately not snapped to the grid — that is what lets them merge across a face — so the placed
  particle may land in a different cell from the one the live per-cell quota counted the leaf in. The
  quota bounds how many leaves are *centred* in a cell, not where each one's particle ends up, so ppc
  there is a target rather than a guarantee. This was already true of the centroid; the new
  placements do not change its nature. A placement that would land inside an embedded boundary falls
  back to the leaf's centroid.
- `kd_placement = sample`/`random` draw from `Random`, so a run using them is reproducible only under
  a fixed seed and rank count, as with `reinitialize`.
- **Options are commented in the repo's style throughout**: a brief `##` comment after each option and
  nothing else. The reorganized `ItoSolver` block carries its grouping as a scope prefix in those
  comments (`kd_cell/kd_patch/kd_amr:`, `kd_amr:`, `nn_amr:`), the way the old
  `nn_pair_tree/nn_pair_onecell/nn_pair_hash:` comments did.
- `.gitignore` gains editor lock/autosave patterns (`.#*`, `#*#`, `*~`), which were breaking `reuse lint`.
- A latent H5Part bug is fixed in passing: `mkdir -p particles/<solver name>` was unquoted, so any
  solver whose name contains a space failed to create its output directory.

### Alternative solutions

- **Two flags, `merge_kernel` (kd/nn/reinitialize) × `merge_scope` (cell/patch/amr).** Rejected: the
  product has holes — there is no nn/patch merge, and `reinitialize` is cell-only — so it would need
  an illegal-combination matrix. One selector with compound names has no invalid states.
- **Keep `kd_split_weight_leaf_dx` alongside a two-valued `kd_partition`.** Rejected: that is the
  status quo, two flags each covering a different subset of the kd methods, and it is the specific
  thing that made the option space hard to navigate.
- **Deprecate the old selectors instead of aborting.** Rejected: each one maps to a method *plus* a
  specifier, so accepting it silently means choosing an axis value on the user's behalf. The abort
  names the replacement.
- **Leave placement to `kd_cell`.** Considered and raised as a concern — the placement interacts with
  the per-cell quota bookkeeping. Overruled deliberately: a leaf landing in a neighbouring cell is
  what the whole-patch scopes are *for*, not a defect.

### Testing

- 2-D and 3-D library builds clean; `pre-commit run --all-files` green.
- Full 2-D `BrownianWalker` suite: all tests, 10 steps, zero aborts, each landing on the intended
  algorithm. `KDAmrSample` runs clean at 1000 steps in 2-D and 10 in 3-D; the `ParticleMerge` example
  runs clean in both dimensions.
- Doc audit: Doxygen clean with warnings-as-errors; a *clean* Sphinx `-W` build (an incremental one
  does not re-read an unchanged `.rst` and hid a real warning); all 181 `literalinclude` directives
  re-checked for excerpts that start or end mid-construct, and six drifted ranges repaired.
- Behaviour: a 500-merge sweep over every method on a 32^2 base grid with two refinement levels and
  8^2 patches, statistics averaged over the last 100 steps. The numbers behind the placement work:

  | configuration | N_eff of 16 | E[(u-1/2)^2] | \|Z\| / shot floor | block % |
  |---|---|---|---|---|
  | `kd_cell` weight centroid | 16.00 | 0.0532 | 62.5 +- 0.1 | 44.4 +- 0.4 |
  | `kd_cell` weight sample | 16.00 | 0.0827 | 4.3 +- 0.8 | 2.7 +- 0.7 |
  | `kd_cell` count sample | 4.66 | 0.0845 | 4.7 +- 0.5 | 2.1 +- 0.6 |
  | `kd_patch` hybrid centroid | 11.40 | 0.0810 | 15.8 +- 0.4 | 2.6 +- 0.7 |
  | `kd_patch` hybrid sample | 12.25 | 0.0845 | 3.3 +- 0.5 | 3.6 +- 1.1 |
  | `kd_amr` carve sample | 12.11 | 0.0837 | 2.8 +- 0.6 | 4.6 +- 0.9 |
  | `reinitialize` (reference) | 16.00 | 0.0833 | 1.8 +- 0.3 | 3.6 +- 0.8 |

  And the distinct-position measurement that motivated `split_placement`, same 500-merge setup, counted
  directly from the particle output:

  | configuration | particles/cell | DISTINCT positions/cell | N_eff |
  |---|---|---|---|
  | `kd_cell` weight, `split_placement = center` | 16 | 1..16 (median **6**) | 16.00 |
  | `kd_cell` weight_capped + jitter | 16 | 15..17 (median **16**) | 16.00 |
  | `kd_cell` count | 16 | 16 | 4.69 |
  | `kd_patch` hybrid | 15..20 | median 15 | 12.26 |
  | `kd_patch` weight_capped + jitter | 16..19 | median 14 | **14.64** |

  Uniform is `E = 0.08333`. Centroid placement contracts the sub-cell distribution and drives the
  particles onto a lattice at 55-63x the shot floor, with 44% peak-to-peak patch-scale density
  modulation on the finest level; `sample` removes both, in every scope. That is what the
  `kd_placement` extension to `kd_patch`/`kd_amr` is for. `kd_partition = count` costs a factor 3.4
  in `N_eff` and buys nothing measurable in return.

### PR-review checklist

Mandatory:

Optional:

https://claude.ai/code/session_012NbqwE8GEeaGoNGuXHxbgZ


rmrsk:
@claude review

rmrsk:
@claude review

rmrsk:
@claude review

PR #722, opened 2026-09-05
Reduce redundant work in the ItoKMC advance, and fix the Krylov convergence scale

rmrsk:
# Summary

Efficiency work on the ItoKMC advance and the paths it leans on. No intended change in results;
one genuine bug fix in the Krylov convergence test.

### Background

Profiling the ItoKMC advance showed time going to work that either did not need doing or did not need
redoing: buffers reallocated every step, a mesh hierarchy built for a code path that never reads it,
per-column divisions in the table lookups that every rate evaluation goes through, and a Krylov exit
test measured against the wrong scale.

### Solution

- **Krylov convergence scale** (`CD_EllipticSolverChainImplem.H`, `CD_KrylovGMRES`, `CD_KrylovBiCGStab`).
  The Krylov solvers tested `eps` against *their own* initial residual, while the hosting solver tests
  against the zero-residual `||rhs - L(0)||`. So `eps` meant different things at the two levels, and the
  better the warm start, the longer the inner solver kept iterating past the point the host was already
  satisfied. The driver now passes `m_residualScale = ||rhs - L(0)||` and the Krylov exit test uses it,
  falling back to the solve's own initial residual when no host supplies one.
- **ItoKMC advance data motion** (`CD_ItoKMCStepper*`, `CD_ItoKMCGodunovStepper*`). Per-species scratch
  (`m_fluidScratchIto`, `m_particleScratchSum`, `m_particleScratchSolid`, `m_particleScratchSumSolid`,
  `m_fluidScratchSolid`) replaces buffers that were allocated and copied per step, and EB emission is
  now answered once, globally, so a step that can emit nothing skips the work entirely.
- **ItoSolver mobility interpolation** (`CD_ItoSolver.H/.cpp`). The `mobility_interp` switch moves out of
  the per-patch dispatch. `velocity` needs a `|v|` hierarchy -- an allocation, a `conservativeAverage`
  and an `interpGhostPwl` per mobile species per step -- which the `direct` path never reads but paid
  for anyway. The per-patch dispatcher goes away with it.
- **LookupTable1D** (`CD_LookupTable1DImplem.H`). One division per row instead of one per column, in
  both interpolation branches and both extrapolation branches. This is the hot path behind every
  tabulated mobility, diffusion coefficient and reaction rate, evaluated per species per cell, and the
  compiler cannot hoist the reciprocal without `-freciprocal-math`.
- **ItoKMC physics scratch** (`CD_ItoKMCPhysics*`, `CD_ItoKMCJSON*`). Thread-local reusable buffers for
  the KMC reaction subsets, propensities and weight partitioning, plus a query for whether the model
  actually reads a species' density gradient so it is not computed when nothing consumes it.
- **`ParticleManagement::partitionParticleWeights`** gains an overload writing into a caller-owned
  buffer, so the callers above stop allocating a vector per call.
- **`LookupTable1D::getIndexLo` out-of-bounds read (pre-existing).** The index was returned unchecked,
  but `interpolate()` reads both `idxLo` and `idxLo + 1`. At `a_x == xmax` the grid formula returns
  `numPoints - 1` by construction, so the upper neighbour was one past the end. Confirmed with
  AddressSanitizer, present on `main`, and affecting uniform and exponential spacing alike; it fired
  only for grids where the division lands exactly on `numPoints - 1` rather than a hair below, which is
  why it went unnoticed. The index is now signed (so a `floor()` below zero clamps instead of wrapping)
  and clamped to `[0, numPoints - 2]`.

### Side-effects

- `ItoSolver::interpolateMobilities(int, const DataIndex&, const EBCellFAB&)` is removed. It was a
  protected per-patch dispatcher with no remaining caller once the switch was hoisted.
- **Breaking interface change.** `ItoKMCPhysics::computeMobilities` and `computeDiffusionCoefficients`
  become out-parameter methods instead of returning `Vector<Real>`. They are pure virtual, so any
  out-of-tree plasma model subclassing `ItoKMCPhysics` must be updated. The in-tree `ItoKMCJSON` is.
- **Results are not bit-for-bit.** Two independent reasons: the Krylov change alters iteration counts,
  and the `LookupTable1D` hoist reassociates floating-point operations (`(a/b)*c` -> `a*(c/b)`), which
  can move the last ULP. `Exec/Tests/tests.py` compares with `h5diff` and no tolerance, so benchmark
  comparisons against a pre-change tree will report differences; they need regenerating.
- `ItoKMCStepper::stepEulerMaruyamaCDR` now skips `conservativeAverage`/`interpGhostPwl` for diffusive
  solvers, relying on `computeDiffusionTermCDR` having already synchronized `phi` earlier in
  `advanceEulerMaruyama`. That ordering invariant is enforced only by comments -- there is no
  synchronization-state mechanism in the codebase to assert it against.
- `Exec/Examples/ItoKMC/AirBasic/example.inputs` turns `ItoKMCGodunovStepper.profile` and
  `ItoKMCJSON.debug` off. Both were left `true` from local profiling runs; `false` is what the class
  `.options` files document as the default, so this only quietens a shipped example.
- Rebased onto `main` after #720. Two conflicts, both in one `@param` block of
  `CD_KDParticleMerge.H`, resolved by keeping #720's parameter name.

### Alternative solutions

- Leaving the Krylov test as-is and lowering `eps` instead: rejected, it hides the mismatch rather than
  fixing it and still penalises good warm starts.
- Allocating the `|v|` hierarchy lazily inside the per-patch dispatch: rejected, the allocation is
  per-hierarchy, not per-patch, so the hoist is the natural place.

### Testing

- 2-D and 3-D `discharge-lib` and `itokmc` build clean.
- `pre-commit run --all-files` green; Doxygen clean with warnings-as-errors; clean Sphinx `-W` build;
  all 181 `literalinclude` directives re-audited for drift after the rebase.
- Regression suite not yet run on the rebased branch -- CI is the check.

### PR-review checklist

Mandatory:

Optional:

https://claude.ai/code/session_012NbqwE8GEeaGoNGuXHxbgZ

### DIFF
--- a/Exec/Examples/ItoKMC/AirBasic/example.inputs
+++ b/Exec/Examples/ItoKMC/AirBasic/example.inputs
@@ -264,7 +264,7 @@ ItoKMCGodunovStepper.secondary_emission                       = after_reactions 
 ItoKMCGodunovStepper.abort_on_failure                         = true               ## Abort on Poisson solver failure or not
 ItoKMCGodunovStepper.abort_max_field                          = 10000              ## Abort if reduced field exceeds this value
 ItoKMCGodunovStepper.redistribute_cdr                         = true               ## Turn on/off reactive redistribution
-ItoKMCGodunovStepper.profile                                  = true               ## Turn on/off run-time profiling
+ItoKMCGodunovStepper.profile                                  = false              ## Turn on/off run-time profiling
 ItoKMCGodunovStepper.plt_vars                                 = none               ## 'conductivity', 'current_density', 'particles_per_patch'
 ItoKMCGodunovStepper.dual_grid                                = false              ## Turn on/off dual-grid functionality
 ItoKMCGodunovStepper.load_balance_fluid                       = false              ## Turn on/off fluid realm load balancing.
@@ -303,7 +303,7 @@ ItoKMCGodunovStepper.reactive_E_centering                     = 0.5             
 # ItoKMCJSON class options
 # ====================================================================================================
 ItoKMCJSON.verbose            = false             ## Turn on/off verbosity
-ItoKMCJSON.debug              = true              ## Turn on/off debugging
+ItoKMCJSON.debug              = false             ## Turn on/off debugging
 ItoKMCJSON.chemistry_file     = chemistry.json    ## Chemistry file
 
 # Kinetic Monte Carlo solver settings.
--- a/Physics/ItoKMC/CD_ItoKMCPhysics.H
+++ b/Physics/ItoKMC/CD_ItoKMCPhysics.H
@@ -271,23 +271,37 @@ public:
 
   /**
    * @brief Compute the Ito solver mobilities.
-   * @param[in] a_time Time
-   * @param[in] a_pos  Position
-   * @param[in] a_E    Electric field
-   * @return Must return a vector of non-negative mobility coefficients for the plasma species
+   * @details Fills a caller-owned buffer rather than returning one. This is called once per grid cell, so a
+   * returned Vector<Real> was a heap allocation and a free per cell -- pure overhead, and (with Chombo memory
+   * tracking compiled in) a pair of counter updates on top. The caller hoists one buffer per patch instead.
+   *
+   * Implementations must size a_mobilities to the number of plasma species and fill every entry; a species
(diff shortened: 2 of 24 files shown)


rmrsk:
@claude review

rmrsk:
@claude review

PR #723, opened 2026-09-06
Select how CDR reaction products reach the mesh

rmrsk:
# Summary

Makes the two halves of a reaction reach the mesh the same way, behind a switch.

### Background

The reaction network handed its two kinds of product to the mesh differently. Ito products became particles at a sub-cell position and reached the mesh through the solver's deposition kernel (CIC by default); CDR products were added straight into the cell that produced them, i.e. NGP by construction. An ionization event creating an Ito electron and a CDR ion therefore spread the electron's charge over a deposition stencil and left the ion's charge in a single cell — a charge separation on the cell scale, proportional to the ionization rate, so largest at the streamer head.

Photoionization showed both branches side by side: `ItoKMCPhysics::reconcilePhotoionization` appends the Ito product and the CDR product at the *same* absorption position, after which the Ito one was deposited with the solver's kernel and the CDR one with a hardcoded NGP.

The same mismatch exists in cut cells, and there it is larger. `ItoSolver.deposition_irreg = mirror` (the setting in `AirBasic` and in the `ItoKMC/JSON` regression test) makes a cut cell hold `n` rather than `kappa*n`. An in-cell CDR injection is `kappa*n` by construction, so in a `kappa = 0.1` cut cell the ionization electron was deposited about ten times higher than its ion partner — right where the sheath forms.

### Solution

New `<stepper>.cdr_products` key:

| value | behaviour |
|---|---|
| `mesh` | Add the production straight into the cell that produced it. The previous behaviour, bit-for-bit. |
| `particle` | Emit it as particles and deposit it with the Ito solvers' deposition. Default. |

`ItoKMCPhysics::reconcileCdrParticles` emits the particles under the same `ParticlePlacement` policy and the same `max_new_particles` budget as the Ito products, and partitions the production the same way — `ParticleManagement::partitionParticleWeights`, so at most `max_new_particles` computational particles of integer weight. The quantization is lossless: `computeReactiveCdrParticlesPerCell` already floors to integers.

`ItoKMCStepper::depositCdrProducts` puts them down through `ItoSolver::depositWeight`, which applies the Ito solvers' deposition *in full* — kernel, coarse-fine strategy, cut-cell strategy, and the mirror or redistribution passes that go with it. That is what makes the two halves agree in cut cells as well as in the bulk. Both kinds of product ride in `m_cdrProducts` — the old `m_cdrPhotoiProducts` generalized from photoionization to all production — so photoionization is no longer a separate path.

Only *production* becomes particles. Mass removed from a CDR species is proportional to the cell-averaged density, which carries no sub-cell information, so it stays in the cell it was taken from; the network's net change is split by sign for this.

Consequences of taking the solver's cut-cell strategy:

- **`redistribute_cdr` is skipped under `cdr_products = particle`.** The reactive redistribution corrects the `kappa` damping that `phi += change/V` imposes on an in-cell cut-cell reaction. A production deposited by the Ito solver carries the solver's normalization instead, so there is no such damping left to correct.
- **`ParticlePlacement::Parent` does not apply to CDR products.** A CDR product has no parents in the cell — its container is transient and holds only what the current step created — so it falls back to a random position. It had been drawing from every Ito species pooled by weight, which in a model with more than one Ito species (the JSON test has two) could place an ion produced by electron-impact ionization on top of a negative ion.

`reconcileCdrParticles` runs *before* the Ito reconciliation rather than after, so both halves read the cell as the reaction network left it. The Ito products cannot move the downstream region — they are placed downstream and `computeUpstreamPosition` compares strictly — but the photoionization products are placed at the absorption position and can, and a cell gaining its first particle of the downstream species switched `hasDownstream` between the two halves.

Also in this PR:

- The placement switch is now one routine, `drawNewParticlePosition()`, shared by the Ito and CDR branches rather than duplicated.
- `Parent` no longer walks the cell's particles once per created particle. `buildParentWeights()` builds the running weight sum once per cell and species; `sampleParentPosition()` binary-searches it. Same accumulation order and same selection rule, so the draw is unchanged.
- **`increment_weights` is removed.** It has not been used, and it silently defeated the placement policy: with it set, Ito production incremented the weights of existing particles while CDR production was placed by the policy, so the two halves diverged again with nothing to warn about.
- `max_new_particles` is now required to be `>= 1`. It divides in the CDR partition and was unvalidated.

### Side-effects

- **Default results change** (`particle` is the default). `cdr_products = mesh` restores the old behaviour exactly.
- **The physics improvement has not been demonstrated.** A fixed-step-count A/B cannot resolve it: over 200 steps of `AirBasic`, the old scheme reaches `t = 1.079e-08` and the new one `9.930e-09` (−8.6%), but the *same* new scheme on a different seed reaches `9.318e-09` (−6.2%). The scheme difference is the size of pure seed-to-seed scatter, and it always will be, because creating CDR product particles consumes random draws and any stochastic ItoKMC model then diverges chaotically. Sizing the real effect needs an ensemble over seeds or a purpose-built deterministic diagnostic. **The −8.6% is not an effect; it is noise.**
- `<stepper>.cdr_products` is parsed with `ParmParse::get`, so it is mandatory. Every input file in the repository has been updated, but an external input deck without the key will abort rather than pick up the default.
- `increment_weights` is no longer read. An old deck that still sets it runs unchanged — `ParmParse` ignores unused keys — but the behaviour is gone.
- `redistribute_cdr` is ignored under `cdr_products = particle`, as above. The `.options` comment says so.
- `ItoKMCStepper::reconcileCdrDensities` changed signature: it takes one *change* field instead of `(new, old)`. It is `virtual`, so a downstream subclass that overrides it needs updating. Nothing in the repository does.
- One deposition and one cross-realm copy per CDR species per step. On `Exec/Examples/ItoKMC/AirBasic` (8 ranks, 200 steps) the wall time was within run-to-run scatter of the pre-change binary — 62.9 s against 59.2 s — measured before the deposition was switched to `ItoSolver::depositWeight`, which adds the mirror pass.
- Photoionization products are deposited under either setting — NGP under `mesh`, with the Ito solvers' deposition under `particle` — since they are created at the photon absorption position and have no in-cell representation to fall back on.
- Placement is stochastic, so the deposited production carries a sampling error falling off as `1/sqrt(N)` in the number of particles used.

### Alternative solutions 

- **A mesh smoothing stencil** instead of particles. Cheaper and noise-free, but it would have to re-implement the EB and coarse-fine handling that `AmrMesh::depositWeight` already provides.
- **Pinning `IrregularDeposition::NGP`** for the CDR products, on the argument that the cut-cell selector is a normalization rather than a spreading rule and that the CDR path is written in the `kappa*n` convention at both ends. This is what the first commit did, and it was wrong twice over: it left the Ito/CDR mismatch in cut cells untouched (see Background), and it made cut cells a one-way sink, since `forceIrregNGP` keys on the particle's *own* cell — a particle in a cut cell exported nothing while its regular neighbours kept depositing into it.
- **Co-locating each CDR product with its Ito partner**, which would cancel a created pair's charge exactly. Rejected as too coupled; the KMC reports only a net per-species change, so the pairing is not actually recorded.
- **A deterministic quadrature** — under CIC, `2^D` particles at the corners of a box of half-width `dx/4` about the cell centre reproduce the expected uniform-placement stencil exactly, with no sampling noise. Not implemented; a plausible third mode for `cdr_products`.

### Testing

Verified on the current branch:

- Builds clean (`Exec/Tests/ItoKMC/JSON`, 2-D, `MPI=TRUE OPT=HIGH`): zero errors, zero warnings.
- `pre-commit run --all-files` passes every hook, `doxygen-check` included.
- `python3 -m sphinx -W --keep-going -b html source build/html` is clean, and the rendered time-step excerpt was checked against the source file by hand.

Verified on the **first** commit, and **not** re-established since — `increment_weights` and the reordered `reconcileCdrParticles` call both sit on these paths:

- `cdr_products = mesh` reproduced the pre-change binary bit-for-bit (`h5diff -c`, zero differences) over 200 steps of `Exec/Examples/ItoKMC/AirBasic` on 8 ranks, `Random.seed = 12345`.
- `cdr_products = particle` with `ItoSolver.deposition = ngp` and centroid placement also reproduced it bit-for-bit.
- `Exec/Tests/ItoKMC/JSON` passed in 2-D and 3-D, no NaNs or aborts.

No simulation has been run since the deposition was switched to `ItoSolver::depositWeight`.

### New types

| type | uses | note |
|---|---|---|
| `ItoKMCStepper::CdrProductInjection` | 5 | The `mesh`/`particle` selector. A domain selector, following `DepositionType` / `IrregularDeposition` / `CoarseFineDeposition` / `ParticlePlacement` in this same code path. |

No new containers: the CDR products reuse the existing `ParticleContainer<NoPayload>` holders, the per-cell production is an `EBAMRCellData`, and the `Parent` running weight sum is a `static thread_local std::vector<Real>` alongside the existing `m_weightScratch`.

### PR-review checklist

Mandatory:

Optional:

https://claude.ai/code/session_01CdEWX5e5x72dU2X3XDdEuw


rmrsk:
@claude review

rmrsk:
@claude review

rmrsk:
@claude review

PR #724, opened 2026-09-07
Report the git version in the startup banner and in HDF5 output

rmrsk:
# Summary

Report the git version of the source tree in the startup banner and in HDF5 output files.

### Background

Output files carried no record of the code that produced them, and the startup banner reported the build configuration but not the version.

### Solution

`Lib/GenerateGitHash.sh` stamps the output of `git describe --tags --always --dirty` into a generated header at build time, and `DischargeIO::gitHash()` returns it. It is printed in the startup banner and written as the `git_hash` attribute of the HDF5 root group of plot and checkpoint files.

Git remains optional: a missing `git`, a source tree that is not a repository, or a repository git refuses to touch all yield `unknown`. The header is rewritten only when the version changes, so a new commit recompiles one object file rather than the library.

### Side-effects

- `Driver::readCheckpointFile` treated every string in the checkpoint header as a realm name, and would have gone looking for a `<version>_loads` data set. It now skips that key; existing checkpoints restart unchanged.
- `tests.py` compares plot files with `h5diff --exclude-attribute /`, since benchmarks are written by a different build than the tree under test and the version therefore always differs. Data sets and per-level attributes are still compared; the root group's `filetype`, `num_levels`, `num_components` and `component_N` names no longer are.
- `git describe` names the newest tag in the *local* clone, so a clone that has not fetched upstream tags reports an older release than it is. The commit hash is exact regardless.

### Alternative solutions

Passing the version as a `-D` flag. Rejected: Chombo's dependency tracking keys on files, never on the values of `-D` flags, so a changed version would silently never trigger a rebuild.

### PR-review checklist

Mandatory:

Optional:

https://claude.ai/code/session_01HNpmtufTUwwBmd5PFqi9uH


rmrsk:
@claude review

PR #725, opened 2026-09-07
Cut the per-super-particle cost of the whole-patch kd merges

rmrsk:
# Summary

Cuts the per-super-particle cost of the whole-patch kd merges (`kd_patch`, `kd_amr`). No intended
change in results.

### Background

`mergeKDCarve()`/`mergeKDInterior()` commit a leaf once per super-particle produced, which makes the
commit loop the busiest loop in the merge. Three things it did per committed leaf were avoidable:

1. **The EB position test ran twice.** `kdPlaceMerged()` returns the centroid unchanged under
   `kd_placement = centroid` (the default), so the centroid fallback was a no-op that re-tested a point
   the previous line had already tested. Both calls evaluate the geometry's implicit function, which is
   the expensive term in this loop.
2. **A patch with no cut or covered cell was tested anyway.** There is no solid in such a patch for a
   merged particle to land in, so the implicit function could only ever return true there.
3. **`payloads`/`weights`/`positions` were declared inside the leaf loop,** costing a malloc/free apiece
   every time a leaf committed. `positions` was also filled unconditionally despite being read only by
   `kd_placement = sample`.

### Solution

Nest the centroid fallback inside the first validity test, so an accepted position costs one predicate
call instead of two.

Add an `isPatchRegular` callback to `mergeKDCarve()`/`mergeKDPatch()`/`mergeKDInterior()`, asked once per
patch and short-circuited into every position test. The merge knows the patch but not the phase; the
solver knows the phase but not which patch a point predicate is being asked about, so a per-patch
callback is the minimal thing that joins them. `ItoSolver` answers it from the realm's `EBISLayout`,
whose `isAllRegular()` is a cached flag on the `EBGraph`.

The callback is defined over the patch **grown by one cell**, because the carve's boundary tier places a
particle built from members up to one ghost cell outside the patch. `EBISBox::isAllRegular()` speaks for
the patch grown by `AmrMesh.eb_ghost`, so the caller confirms that option reaches at least one cell
rather than assuming it; below one, no patch reports regular and every position is tested as before.

Hoist the commit scratch next to the existing `leaves`/`members` scratch and `clear()` it per leaf, so
capacity carries across leaves and patches. Fill `positions` only for `kd_placement = sample`;
`kdPlaceMerged()` now documents that it may be empty otherwise and asserts the pairing in the branch
that reads it. The carve boundary tier was also missing the `positions.reserve()` its interior tier had.

### Side-effects

**API.** `mergeKDCarve()`, `mergeKDPatch()` and `mergeKDInterior()` take one more template parameter
(`PatchRegular`) and one more argument (`a_isPatchRegular`). All three call sites in `ItoSolver` are
updated: two in `mergeKDImpl()` and one in `mergeKDSkinNn()`. No new types, structs, containers or maps
are introduced — `PatchRegular` is a template parameter, and the callback is a lambda at each call site.

**Bit-identity.** The fallback nesting and the scratch hoisting are bit-identical by construction: the
same values are accumulated in the same order, the predicate is pure so dropping a redundant call has no
side effect, and RNG draws are untouched (`Random::getUniformReal01()` appears only in `kdPlaceMerged()`
and `kdSplitDisplace()`, both invoked unconditionally before any validity test).

Skipping the predicate on regular patches is bit-identical **iff** EBIS regularity agrees with the sign
of the implicit function — the assumption the EB discretization already rests on. It would diverge only
for a solid feature small enough that EBIS resolves no irregular cell for it, in which case a leaf that
was previously rejected now commits. That would not be a subtle difference: the newly committed leaf also
consumes an `a_allocateID()`, shifting every subsequent id on the rank and changing particle population
and ordering, so a benchmark comparison would show it loudly.

**Testing status.** This branch compiles clean (`Exec/Examples/ItoKMC/AirBasic`, DIM=2, MPI, HDF5,
OPT=HIGH, no warnings in the touched files). The regression suite has **not** been run and no benchmark
comparison has been made, so the bit-identity argument above is static reasoning only.

**Unrelated finding, not addressed here.** `kd_partition = weight` maps to a crossover of
`Real::max()`, which makes `kdSplitWeightMedian()` — a full `std::sort` — run at every node of the tree
rather than only at sub-cell-width nodes, taking `buildKDQuotaLeaves()` from O(N log N) to O(N log²N).
`kd_partition = count` and `hybrid` are unaffected. A weighted selection would remove it; that is a
separate change.

### Alternative solutions

**Bake regularity into `isPositionValid` in the solver.** Rejected: the predicate is a point function and
cannot tell which patch it is being called from, so it would have to re-derive the patch from the
position — work the merge algorithm has already done.

**Give `ItoSolver` a private `isPatchAllRegular()` member** instead of repeating the small lambda in
`mergeKDImpl()` and `mergeKDSkinNn()`. Rejected in favour of the existing house pattern of a short
cross-reference comment (as `mergeNnPairSearch()`'s id-allocator rationale already does), which adds no
new class surface.

**Hoist the `patchRegular` branch out of the commit loop** by specializing the loop. Rejected: it is a
perfectly predicted branch on a loop-invariant bool, and duplicating the commit loop to remove it is not
worth the divergence risk.

**Replace `std::sort` in `kdSplitWeightMedian()` with a weighted selection.** Out of scope here; see the
note under Side-effects.

### PR-review checklist

Mandatory:

Optional:

https://claude.ai/code/session_01Jtr7vNPRQMyVgCY2QZFKMJ

### DIFF
--- a/Source/ItoDiffusion/CD_ItoSolver.cpp
+++ b/Source/ItoDiffusion/CD_ItoSolver.cpp
@@ -21,6 +21,7 @@
 #include <ParmParse.H>
 #include <ParticleIO.H>
 #include <EBCellFactory.H>
+#include <EBISLayout.H>
 
 // Our includes
 #include <CD_ItoSolver.H>
@@ -4687,6 +4688,16 @@ ItoSolver::mergeKDSkinNn(ParticleContainer<ItoMergeParticle>& merge, const Vecto
     return implicitFunction->value(a_pos) < 0.0;
   };
 
+  // See mergeKDImpl() for why the whole-patch kd merges get a per-patch escape hatch from
+  // isPositionValid(), and for why it is the patch grown by one cell that has to be regular.
+  const Vector<EBISLayout>& ebisLayout = m_amr->getEBISLayout(m_realm, m_phase);
+
+  const bool ebGhostCoversHalo = m_amr->getNumberOfEbGhostCells() >= 1;
+
+  auto isPatchRegular = [&ebisLayout, ebGhostCoversHalo](const int a_lvl, const DataIndex& a_din) -> bool {
+    return ebGhostCoversHalo && ebisLayout[a_lvl][a_din].isAllRegular();
+  };
+
   // See mergeNnPairSearch() for the id-allocator rationale.
   constexpr ParticleID rankStride = 1000000000000LL;
   const ParticleID     rankBase   = static_cast<ParticleID>(procID()) * rankStride;
@@ -4748,7 +4759,8 @@ ItoSolver::mergeKDSkinNn(ParticleContainer<ItoMergeParticle>& merge, const Vecto
                                                               kdCombine,
                                                               kdScatter,
                                                               allocateID,
-                                                              isPositionValid);
+                                                              isPositionValid,
+                                                              isPatchRegular);
 
   merge.clearGhostParticles();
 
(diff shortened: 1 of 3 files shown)


rmrsk:
@claude review

PR #726, opened 2026-09-07
Filter cut cells consistently in DataOps::filterSmooth

rmrsk:
# Summary

### Background

The cut-cell kernel in `DataOps::filterSmooth` was never written. The irregular VoF iterator was
built per patch and then discarded, the loop that would have used it was commented out (and named an
`irregularKernel` that did not exist anywhere in the function), and the regular kernel ran over the
whole valid box. Cut cells were therefore filtered with the plain Cartesian tensor stencil, reading
covered neighbours as zeros. A comment claimed those cells were not filtered; nothing implemented
that.

For a cell flush against a planar embedded boundary the covered part of the stencil carries a weight
of 0.25 — in 2D one face at `alpha*beta` = 0.125 plus two corners at `beta^2` = 0.0625, and in 3D one
face, four edges and four corners summing to the same. Such a cell therefore lost roughly a quarter
of its value per pass, biased systematically towards the boundary and compounding with the number of
passes. Multiply-cut cells kept stale values in the irregular part of the `EBCellFAB` entirely, since
the regular kernel only ever writes the `FArrayBox`.

Separately, `CH_assert(a_alpha <= 0.0)` fired for every valid input, so any DEBUG build with
filtering enabled aborted. This went unnoticed because every `filter_num` default is zero.

The routine filters the space charge and the semi-implicit conductivity in
`CdrPlasmaGodunovStepper` and `ItoKMCGodunovStepper`.

### Solution

Filter cut cells with the same tensor-product weights, but drop neighbours that are covered, outside
the domain, or unreachable by a monotone path, and place their weight on the centre cell.

Monotone-path reachability is symmetric — if `j` is reachable from `i` then `i` is reachable from
`j` — so two cells always assign each other the same weight. Moving the dropped weight onto the
centre makes the rows sum to one, and the symmetry then makes the columns sum to one as well. The
first leaves a uniform field uniform; the second conserves `sum(phi)`, which is the condition for the
effective filtered deposition shape function to remain a partition of unity. Dropping a neighbour
without moving its weight to the centre would break both, and filling it with a covered value drains
`phi` into the EB. The stencil is convex for `alpha` in [0, 1], so it cannot manufacture the
grid-scale oscillation it exists to remove.

The set handled by the cut-cell kernel is grown by the full stride rather than by `stride - 1`. A
regular cell can see a covered cell at a corner of its stencil, and the cells between the two are
necessarily irregular, so the grown set is exactly the set of cells whose stencil can reach a covered
cell.

The `CH_assert(a_alpha <= 0.0)` is removed rather than corrected to `<= 1.0`. The bound is not simply
mistyped: the compensating pass in `CdrPlasmaGodunovStepper` deliberately runs with
`alpha = 1 + n/2`, so there is no upper bound to assert.

No new types or containers are introduced.

### Side-effects

- Results change near embedded boundaries wherever filtering is enabled; that is the intent. Regular
  cells away from the EB are untouched, and since every `filter_num` default is zero, all default
  configurations and all benchmark files are bit-for-bit unchanged.
- The cut-cell path is more expensive than the regular kernel. The grown `IntVectSet` and its
  `VoFIterator` are rebuilt per patch per call, because `EBISBox::getIrregIVS` returns by value and
  the grown set does not match the iterator prebuilt in `PhaseRealm`; and `VofUtils::getVofsInRadius`
  walks the graph and allocates once per cut cell per component. The affected set is a stride-thick
  surface band. This is not measured — `CH_TIME("DataOps::filterSmooth(LevelData<EBCellFAB>)")`
  already reports it. If it proves to matter, a grown-irregular iterator on `PhaseRealm` (following
  `getFaceIteratorWithTangentialGhosts`) removes the per-call rebuild, and cached stencils in the
  manner of `EBNonConservativeDivergence` remove the graph walk; the stencil topology depends only on
  the stride, not on `alpha`, so one cached set serves both the smoothing and compensating passes.
- `a_zeroEB` no longer affects the filtered result in cut cells, since covered neighbours are now
  excluded structurally rather than by being zeroed. The parameter and its clone-zeroing behaviour
  are retained so the signature and the five call sites are unchanged.
- This branch has not been compiled or run. `pre-commit` (clang-format, reuse, codespell, cppcheck,
  doxygen) passes, and `python3 -m sphinx -W --keep-going` builds clean; no RST file references
  `DataOps`, so no `literalinclude` line references can have drifted.

### Alternative solutions 

- **Conservative flux form with aperture weighting.** Cast the filter as one explicit diffusion step,
  `kappa*rho += beta*sum_f area_f*(rho_nbr - rho)`, conserving `sum(kappa*rho)` rather than
  `sum(rho)` and taking zero flux through the EB face, with the `kappa -> 0` overshoot handled by
  hybrid divergence plus redistribution exactly as `CdrSolver` does for diffusion. Rejected as
  disproportionate to the defect: it requires stateful objects that a static `DataOps` cannot own,
  and the `1/kappa` it introduces is a new small-cell instability that the stencil in this PR does
  not have.
- **Skipping whole directions in which a neighbour is covered.** Simpler, and it does preserve
  constants, but the decision is per cell and therefore asymmetric: cell `i` skips a direction that
  its neighbour `j` does not, so `j` hands charge to `i` that `i` never returns, the columns stop
  summing to one, and charge drifts with a systematic geometric sign. Deciding reachability per
  neighbour pair rather than per cell is precisely what makes the stencil here symmetric.
- **Genuinely skipping cut cells**, as the stale comment claimed. Also not conservative: regular
  cells would still draw from cut cells that do not reciprocate.

### PR-review checklist

Mandatory:

Optional:

https://claude.ai/code/session_01J7VBPPpSHBXhyo99QUSeFL

### DIFF
--- a/Source/Utilities/CD_DataOps.H
+++ b/Source/Utilities/CD_DataOps.H
@@ -287,8 +287,12 @@ public:
    * @param[in]    a_stride Cell stride s
    * @param[in]    a_zeroEB Set value in covered cells to zero or not
    * @note This routine will fill ghost cells outside the domain boundary, but ignore the corner ghost cells at the
-   * domain corners. This routine may be dangerous to use with EBs as the data inside the EB is either kept as-is or
-   * replaced by zero.
+   * domain corners.
+   * @note Cut cells are filtered with the same tensor-product weights, but neighbors that are covered, outside the
+   * domain, or unreachable by a monotone path are dropped from the stencil and their weight is placed on the center
+   * cell. The resulting stencil is symmetric, so the filter conserves sum(phi) and leaves a uniform field uniform.
+   * @note a_alpha may exceed one, which turns the filter into the compensating (sharpening) pass used to restore the
+   * low-wavenumber amplitude after a sequence of smoothing passes.
    */
   static void
   filterSmooth(EBAMRCellData& a_data, const Real a_alpha, const int a_stride, const bool a_zeroEB) noexcept;
@@ -301,8 +305,12 @@ public:
    * @param[in]    a_stride Cell stride s
    * @param[in]    a_zeroEB Set value in covered cells to zero or not
    * @note This routine will fill ghost cells outside the domain boundary, but ignore the corner ghost cells at the
-   * domain corners. This routine may be dangerous to use with EBs as the data inside the EB is either kept as-is or
-   * replaced by zero.
+   * domain corners.
+   * @note Cut cells are filtered with the same tensor-product weights, but neighbors that are covered, outside the
+   * domain, or unreachable by a monotone path are dropped from the stencil and their weight is placed on the center
+   * cell. The resulting stencil is symmetric, so the filter conserves sum(phi) and leaves a uniform field uniform.
+   * @note a_alpha may exceed one, which turns the filter into the compensating (sharpening) pass used to restore the
+   * low-wavenumber amplitude after a sequence of smoothing passes.
    */
   static void
   filterSmooth(LevelData<EBCellFAB>& a_data, const Real a_alpha, const int a_stride, const bool a_zeroEB) noexcept;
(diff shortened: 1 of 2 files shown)


rmrsk:
@claude review

PR #727, opened 2026-09-08
Speed up EBIS geometry generation

rmrsk:
# Summary

Speeds up EBIS geometry generation. The Armadillo 3-D example builds its geometry **2.62x** faster
(4.67 s -> 1.78 s, 8 ranks), and **~3.2x** with an opt-in flag added here. Every change is
bit-identical: the whole `Exec/Tests/Geometry` suite passes exact `h5diff` against pre-change
benchmarks in 2-D and 3-D, 38 comparisons, zero differences.

Most of the work is in the Chombo submodule (chombo-discharge/Chombo-3.3#29). This PR adds one
chombo-discharge change and the five submodule bumps.

### Background

Profiling `Exec/Examples/Electrostatics/Armadillo` in 3-D with `Driver.geometry_only = true`
(128^3 finest EBIS domain, 7 EBIS levels, 8 ranks) showed the geometry build was dominated by work
that had already been done once.

Two examples were used throughout, because they stress opposite paths:

- **Armadillo** — an STL tessellation through EBGeometry's BVH, so `BaseIF::value` is expensive.
  Implicit-function bound: `ScanShop::makeGrids` 34%, `GeometryShop::fillGraph` 48%.
- **ProfiledSurface** — analytic CSG with an electrode and a dielectric, so `value` is cheap and two
  EBIS's are built. Algebra and communication bound: cut-cell moments 58%, coarsening cascade 31%.

Measuring only one is misleading, and the 2-D regression tests are more misleading still: there the
coarsening cascade is 60%+ of the build, whereas in a real 3-D case it is 3%, because their finest
level is trivial and fixed per-level costs dominate. Several items that look large in the 2-D
profiles were measured and dropped for this reason.

### Solution

The chombo-discharge change is a new option, `ScanShop.distance_skip`, **off by default**.

`ScanShop::isRegular` and `isCovered` walk every cell centre of the grown box and compare the
implicit function against the cell half-diagonal. That test is already a distance argument -- it is
only sound if the implicit function is a signed distance function -- so the same assumption lets the
scan pass over cells: when a point clears the threshold by a slack `s`, no cell centre closer than
`s` can cross it, so `floor(s/dx)` cells along the scan line can be skipped.

On Armadillo this takes `ScanShop::isCovered` from 0.66 s to essentially nothing, and the geometry
build from 2.21 s to 1.57 s. The cost there is entirely *confirming that covered boxes are covered*,
which otherwise has to visit all 8000 cells of each grown box.

The submodule bumps carry five Chombo commits: a corner-value cache in `GeometryShop::fillGraph`, a
cut-edge intersection cache, a blocked `LSquares` matrix allocation, cached moment exponent lists,
and shared exchanges between the two `EBISLevel` coarsening steps. Details in the Chombo PR.

### Side-effects

**`ScanShop.distance_skip` is off by default and I recommend it stays off until the contract is
documented.** Not every implicit function here is a signed distance function:

- `PerlinPlaneSdf::value` returns `plane->value(x) + perlin->value(xp)`, whose gradient magnitude
  exceeds one wherever the noise has a gradient. Tested anyway: 8 configurations of `NoisePlane`
  and `RoughSphere`, including resolved noise with local slopes to 6x and amplitude 60x dx, all
  bit-identical while the skip fired hard (`isCovered` 0.037 s -> 0.003 s). There is a structural
  reason -- the plane term is constant along the scan line, so the excursion is bounded by the
  noise amplitude.
- `HyperboloidOneIF` and `HyperboloidTwoIF` return `sum(+-(x-c)^2/r^2) -+ 1`, a dimensionless
  quadratic form rather than a length, with gradient `2(x-c)/r^2`. Nothing in the repository uses
  them, but they are public API, and `BaseIF` makes no distance contract at all. For such a
  function the *existing* half-diagonal test is already meaningless; the skip widens that gap from
  a per-cell error to a pruning error over a span.

Turning the default on would be worth roughly 20% of the Armadillo build, but it should come with
"ScanShop requires a signed distance function" written into the class docs and the Sphinx geometry
page, so the requirement is discoverable. Happy to do that in this PR if you want it.

Also in this PR: `CD_ScanShopImplem.H` carried a stray `@return` inside its SPDX block and a
duplicated `@brief` instead of `@file`. Doxygen was attaching that stray tag to whichever function
came first in the file, which is the only reason `ScanShop::isRegular` ever passed the doxygen check
without documenting its own return value. Adding a function ahead of it exposed this; the header is
repaired and `isRegular` now documents its return.

**No new types or containers are introduced**, in either repository. The caches added on the Chombo
side use `BaseFab<Real>`.

### Alternative solutions

- **Defaulting `distance_skip` on.** See above -- deferred to a decision about documenting the SDF
  contract, not to a technical obstacle.
- **A Lipschitz constant parameter** instead of assuming 1. Rejected: nobody knows their constant,
  and a wrong value fails just as silently.
- **Evaluating the whole grown box into an array** rather than skipping. Rejected: it loses the
  early exit, which is what makes the near-surface boxes cheap.
- On the Chombo side, a bit-changing `LSquares` variant was measured (a further 1.17x on
  ProfiledSurface) and deliberately left out because it perturbs the last bits; details in the
  Chombo PR.

### PR-review checklist

Mandatory:

Optional:

---

**Verification.** `Exec/Tests/Geometry`, benchmarks generated on the pre-change tree, compared
against this branch, `make pristine` on both sides, `-cores 12` throughout:

| | tests | comparisons | mismatches |
|---|---|---|---|
| 2-D | 21 | 19 | **0** |
| 3-D | 20 | 19 | **0** |

The three non-comparisons are `Cylinder` in both dimensions and `RodPlaneProfile2d`, which write no
HDF5 at all (`numOutputComp == 0` in `Driver::writePlotFile`); they still ran as crash checks.
Because these tests use `GeometryStepper`, which contributes no plot variables, the compared files
hold only the EB moment data -- volume fractions, boundary areas, all area fractions, EB normals and
the derived distance -- with no solver in the path to amplify or mask a difference.

### DIFF
--- a/Exec/Examples/Electrostatics/Armadillo/armadillo3d.inputs
+++ b/Exec/Examples/Electrostatics/Armadillo/armadillo3d.inputs
@@ -126,7 +126,7 @@ Tessellation.mesh_file       = armadillo.stl    ## Ply file
 Tessellation.partitioner     = default          ## Partitioner. Valid options are 'default', 'overlap', 'sah'
 Tessellation.use_bvh         = true             ## Use bounding volume hierarchy for accelerating search
 Tessellation.z_coord         = 0.0              ## z-coordinate for 2D slicing this object
-Tessellation.flip_inside     = false
+Tessellation.flip_inside     = true
 
 # ====================================================================================================
 # FIELD_STEPPER CLASS OPTIONS
--- a/Source/Geometry/CD_ScanShop.H
+++ b/Source/Geometry/CD_ScanShop.H
@@ -172,6 +172,13 @@ protected:
   bool m_profile;
 
   /**
+   * @brief Whether or not the box scan may skip cells using the implicit function value as a distance.
+   * @details Only correct when the implicit function is a signed distance function. Set through
+   * ScanShop.distance_skip.
+   */
+  bool m_distanceSkip;
+
+  /**
    * @brief Scan level where we first begin to break up boxes. This is relative the EBIS level.
    */
   int m_scanLevel;
@@ -255,10 +262,24 @@ protected:
   buildCoarseLevel(int a_level, int a_maxGridSize);
 
   /**
+   * @brief Number of cells along the scan line that can be passed over without evaluating the implicit function.
+   * @details A signed distance function changes by at most the distance travelled, so a cell centre closer than
+   * a_slack to the last evaluated point cannot cross the regular/covered threshold. Returns zero unless the
+   * distance skip is enabled, because implicit functions that are not signed distance functions (noise-displaced
+   * surfaces, for example) can change faster than that.
+   * @param[in] a_slack Distance by which the last evaluated point cleared the threshold
+   * @param[in] a_dx    Grid resolution on level
+   * @return Number of additional cells to advance the scan by
+   */
+  inline int
+  getScanSkip(Real a_slack, Real a_dx) const;
+
+  /**
    * @brief Check if every point in input box is regular
    * @param[in] a_box    Cell-centered box
    * @param[in] a_probLo Lower-left corner of simulation domain
    * @param[in] a_dx     Grid resolution
+   * @return True if regular, false otherwise
    */
   inline bool
   isRegular(Box a_box, const RealVect& a_probLo, Real a_dx) const;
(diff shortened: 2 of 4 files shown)


rmrsk:
@claude review

rmrsk:
@claude review

PR #729, opened 2026-09-11
Fix reaction loss in the hybrid tau-leaping algorithm for stiff chemistries

### LINKED ISSUE #728: KMC: hybrid tau-leaping loses reactions for stiff chemistries
## Summary

The hybrid tau-leaping algorithms in `KMCSolver` lose reactions when the chemistry is stiff, i.e.
when one reaction turns its reactant over faster than the step. With default settings the mean
population comes out **7 % low** on the small test problem below, and a change to `crit_num` that
should be harmless takes it to **66 % low**.

There appear to be two independent causes. The second one looks like a plain bug rather than a
resolution issue.

Found while comparing `ItoKMCJSON` against a particle Monte Carlo model (`particle_swarm`) for
electron avalanches, on a two-group electron chemistry where a fast relaxation reaction feeds a slow
branching reaction. Version `v26.09-17-g05715d723`. A ready-to-build reproducer is linked below.

## Reproducer

Ready to build, on the `kmc-stiff-repro` branch of my fork:

**[`Exec/Examples/KineticMonteCarlo/StiffTwoGroup`](https://github.com/rmrsk/chombo-discharge/tree/kmc-stiff-repro/Exec/Examples/KineticMonteCarlo/StiffTwoGroup)**

```
git remote add rmrsk https://github.com/rmrsk/chombo-discharge
git fetch rmrsk kmc-stiff-repro && git checkout rmrsk/kmc-stiff-repro
cd Exec/Examples/KineticMonteCarlo/StiffTwoGroup
make -s -j<num_proc> OPT=HIGH DEBUG=FALSE DIM=2 main
mpirun -np 12 ./main2d.*.ex example.inputs
```

The branch is cut from `main` (`0fd5c9c17`) and adds nothing but that directory. The numbers below
were taken at `v26.09-17-g05715d723`, but `Source/KineticMonteCarlo` and `Physics/ItoKMC` are
identical between the two, so they reproduce on either.

The files are inlined below as well. Constant rates, no data files, no field. Two species, `e` (slow, branching) and
`el` (fast, relaxes back), plus an inert product.

```
  e  -> el + el + M+     nu_i = 7.468e9  1/s
  el -> e                nu_r = 1.741e11 1/s     (23x faster)
```

Starting from one `e`, the total `e + el` grows at the dominant eigenvalue of
`[-nu_i, nu_r; 2 nu_i, -nu_r]` = `6.8987e9 1/s`, and the exact mean at `t = 1 ns` is
**993.89** (from the moment equations, which close because every reaction is first order).

<details>
<summary><code>chemistry.json</code></summary>

```json
{
    "gas": {
        "background species": [
            { "id": "M", "molar fraction": { "type": "constant", "value": 1.0 } }
        ],
        "law": {
            "id": "ideal_gas",
            "plot": false,
            "ideal_gas": { "type": "ideal", "temperature": 300, "pressure": 0.233E5 }
        }
    },
    "particle placement": { "method": "random", "species": "e" },
    "alpha": { "type": "constant", "value": 0.0 },
    "eta":   { "type": "constant", "value": 0.0 },
    "plasma species": [
        { "id": "e",  "Z": -1, "solver": "ito", "mobile": false, "diffusive": false, "initial density": 0.0 },
        { "id": "el", "Z": -1, "solver": "ito", "mobile": false, "diffusive": false, "initial density": 0.0 },
        { "id": "M+", "Z":  1, "solver": "cdr", "mobile": false, "diffusive": false, "initial density": 0.0 }
    ],
    "photon species": [],
    "plasma reactions": [
        { "reaction": "e -> el + el + M+", "type": "constant", "value": 7.468E9,  "include_dt_calc": true },
        { "reaction": "el -> e",           "type": "constant", "value": 1.741E11, "include_dt_calc": true }
    ],
    "photoionization": []
}
```
</details>

<details>
<summary><code>minimal.cpp</code></summary>

```cpp
#include <limits>
#include "ParmParse.H"
#include <CD_ItoKMCJSON.H>

using namespace ChomboDischarge;
using namespace Physics::ItoKMC;

int
main(int argc, char* argv[])
{
#ifdef CH_MPI
  MPI_Init(&argc, &argv);
#endif
  ParmParse pp(argc - 2, argv + 2, NULL, argv[1]);

  int  numRuns;
  Real stopTime, maxDt;
  pp.get("num_runs", numRuns);
  pp.get("stop_time", stopTime);
  pp.get("max_dt", maxDt);

  Random::seed();

  auto      physics   = RefCountedPtr<ItoKMCJSON>(new ItoKMCJSON());
  const int numPlasma = physics->getNumPlasmaSpecies();
  const int numPhoton = physics->getNumPhotonSpecies();

  Vector<Real>     particles(numPlasma, 0.0), photons(numPhoton, 0.0), phi(numPlasma, 0.0);
  Vector<RealVect> gradPhi(numPlasma, RealVect::Zero);

  physics->defineKMC();

  Real sum = 0.0;
  for (int run = 0; run < numRuns; run++) {
    for (int i = 0; i < numPlasma; i++) {
      particles[i] = 0.0;
    }
    particles[0] = 1.0; // one electron in the reacting group

    for (Real t = 0.0; t < stopTime;) {
      const Real dt    = std::min(maxDt, stopTime - t);
      Real       newDt = std::numeric_limits<Real>::max();

      physics->advanceKMC(particles, photons, newDt, phi, gradPhi, dt,
                          RealVect::Zero, RealVect::Zero, 1.0, 1.0);
      t += dt;
    }
    sum += particles[0] + particles[1]; // e + el
  }

  pout() << "mean total electron number = " << sum / numRuns << endl;

#ifdef CH_MPI
  MPI_Finalize();
#endif
  return 0;
}
```
</details>

<details>
<summary><code>repro.inputs</code></summary>

```
ItoKMCJSON.verbose            = false
ItoKMCJSON.debug              = false
ItoKMCJSON.chemistry_file     = chemistry.json
ItoKMCJSON.max_new_particles  = 16
ItoKMCJSON.max_new_photons    = 64
ItoKMCJSON.increment_weights  = false
ItoKMCJSON.crit_num           = 5
ItoKMCJSON.prop_eps           = 0.5
ItoKMCJSON.SSA_num            = 5
ItoKMCJSON.SSA_lim            = 2.0
ItoKMCJSON.max_iter           = 5
ItoKMCJSON.exit_tolerance     = 1.E-6
ItoKMCJSON.algorithm          = hybrid_midpoint

Random.seed = -1
num_runs    = 20000
stop_time   = 1.0E-9
max_dt      = 1E-11
```
</details>

Results, 12 ranks x 20000 realizations = 240k, so the sampling error on the mean is about 0.2 %:

| configuration | mean at 1 ns | vs exact 993.89 |
|---|---|---|
| `algorithm=ssa` | 994.93 | +0.10 % |
| `hybrid_midpoint`, `max_dt=1e-11` (as above) | 922.29 | **-7.20 %** |
| `hybrid_midpoint`, `max_dt=1e-12` | 995.08 | +0.12 % |
| `hybrid_midpoint`, `max_dt=1e-11`, `crit_num=500` | 336.36 | **-66.16 %** |
| ... same but `SSA_lim=0` | 992.21 | -0.17 % |
| ... same but `SSA_lim=1000` | 996.55 | +0.27 % |

SSA is correct, so the reference is not in doubt.

## Problem 1: the SSA/critical branch is chosen from a random draw

In `KMCSolver::advanceHybrid`:

```cpp
const Real curDt = std::min(a_dt - curTime, std::min(dtCrit, dtNonCrit));
...
const bool useSSA = (m_numSSA >= one) && (A * curDt < m_SSAlim);
```

`dtCrit` is drawn as the waiting time to the next *critical* firing, so `curDt` can be a random
exponential, and `useSSA` then selects the branch from the realized value of that draw.

* In the critical branch the pending critical reaction is applied
  (`stepSSA(state, criticalReactions, propensitiesCrit)`).
* In the SSA branch it is not. That branch only fires what its own independent draws produce, and
  when `dtSSA + dtReact >= curDt` it sets `dtSSA = curDt` and advances to the end of the substep
  having fired nothing.

So whenever `curDt == dtCrit` and the SSA branch is taken, the firing that `dtCrit` was drawn for is
silently dropped. Short waiting times preferentially satisfy `A * curDt < SSA_lim`, so it is exactly
the short intervals that lose their reaction.

The two pure regimes are both correct and the mixture is not, which is what the last three rows of
the table show: `SSA_lim = 0` (never SSA) gives -0.17 %, `SSA_lim = 1000` (always SSA) gives
+0.27 %, and the default `SSA_lim = 2` gives -66 %. `crit_num = 5` normally hides this because few
reactions are ever critical; raising it routes almost every firing through the broken path.

Either of these should fix it: decide the branch before drawing `dtCrit` (or from a quantity that
does not depend on it), or apply the pending critical firing in the SSA branch as well.

## Problem 2: nothing bounds per-reaction turnover

With `max_dt = 1e-11` we have `nu_r * dt = 1.74`, so the relaxation reaction is drawn
`Poisson(nu_r * n_el * dt)` = `Poisson(133)` firings against a population of `n_el = 76`. The
resulting state is invalid, `advanceHybrid` rejects it and halves `dtNonCrit`, and the accepted step
is then conditioned on the random draw - a biased retry. The bias scales with the step and vanishes
below roughly `1 / (5 nu_r)`.

Neither step control catches it, and for the same reason: both bound the **net** change per species.
For a fast intermediate in quasi-steady state, production and loss nearly cancel - here
`2 nu_i n_e = 1.380e10 N` against `nu_r n_el = 1.328e10 N`, a net 26x smaller than either term - so:

* `KMCSolver::computeDt` (the physics-level limit, mean term only) returns 146 ps;
* `KMCSolver::getNonCriticalDt` does carry the variance term `dt2 = f^2 / sigma^2`, but it also
  bounds only the change in `n_el`, which is genuinely small however many times the group turns
  over. It permits 35.6 ps here.

What neither bounds is `a_j * tau` against the reactant population, i.e. the Cao-Gillespie-Petzold
criticality quantity used as a *step* constraint rather than only as a partition test. Adding
`a_j * tau <= eps * L_j` would have given `tau <= 2.9 ps` at `eps = 0.5`.

Related: `partitionReactions` compares the number of firings to depletion against a fixed
`crit_num`, with no reference to the rate. A reaction that will turn its reactant over several times
within the step is classed non-critical whenever the population happens to exceed 5.

Implicit tau-leaping does not rescue this. At `max_dt = 1e-11` on the same problem,
`hybrid_implicit_euler` is stable but lands **+7.2 %** high; `hybrid_explicit_euler` -7.1 %,
`hybrid_midpoint` -3.8 %, `hybrid_prc` -4.3 % (100k realizations, `prop_eps = 0.5`).

## Not explained

Tightening `prop_eps` at `max_dt = 1e-11` makes the bias *worse* rather than better, and it does so
with `SSA_lim = 0` too, so it is not Problem 1 (100k realizations, `crit_num = 5`):

| `prop_eps` | `SSA_lim = 0` | `SSA_lim = 2` |
|---|---|---|
| 1.E99 | -1.53 % | -2.10 % |
| 0.5 | -3.17 % | -3.97 % |
| 0.1 | -3.72 % | -5.16 % |

My guess is the rejection loop in Problem 2 - more substeps, more rejections, each one conditioning
the step on the random draw - but I did not confirm it. The trend disappears once the step resolves
the fast reaction.

## Suggestion

Independently of the fixes, it would help a lot if the solver warned when
`max_j (a_j * dt / L_j)` exceeds ~1 for any reaction. Both problems here are silent: the results
look plausible, they converge under refinement, and nothing in the output suggests the step was
inadequate. The chemistry that triggered this is only mildly stiff by combustion or biochemical
standards (a rate ratio of 23), so it is probably not an unusual situation.

rmrsk:
# Summary

The hybrid tau-leaping algorithms in `KMCSolver` lost reactions when one reaction turns its reactant over faster than the step. On the two-group reproducer from the issue the mean population came out 7 % low with default settings and 66 % low with `crit_num = 500`. This PR fixes the three underlying defects in `KMCSolver::advanceHybrid` / `KMCSolver::getNonCriticalTimeStep`, adds the reproducer as convergence test C3, and updates the KMC documentation.

Closes #728 

### Background

The issue reports two independent causes, found while comparing `ItoKMCJSON` against a particle swarm model on a two-group chemistry `e -> el + el + M+` (slow) and `el -> e` (23x faster):

1. `advanceHybrid` chose the SSA fallback from `A * curDt < SSA_lim` where `curDt` included the exponential waiting time `dtCrit` to the next critical firing. When the draw was short enough to select the SSA branch, that branch advanced from its own independent draws and the firing `dtCrit` had been drawn for was dropped. Short waiting times preferentially satisfy the SSA test, so exactly the short intervals lost their reaction.
2. Both leap conditions bound the *net* change of each reactant. For a fast intermediate in quasi-steady state production and loss nearly cancel, so nothing bounded the per-reaction turnover; the Poisson draws of the fast reaction routinely exceeded the population, the step was rejected and halved, and the accepted step was conditioned on the random draw.

The issue also had a "not explained" table: tightening `prop_eps` made the bias *worse*, even with `SSA_lim = 0`. Measuring after fixing 1 and 2 left a residual −3 % at default settings that grew to −7 % with `SSA_lim = 0`, and it turned out to be a third deviation from Cao, Gillespie & Petzold (2006): the leap condition gathered reactant species from the *non-critical* reactions only and skipped any species at zero population. A reactant of a critical reaction that the leap produces from nothing (here `el` when `n_el = 0`) was therefore not bounded at all. The leap produced such particles and left them idle for its whole length while their fast consumption stayed frozen at zero propensity — a delay on every relaxation, hence a growth-rate bias. Because the floor `f = max(eps X, 1)` is what governs this regime, the delay is independent of `prop_eps`, which is why tightening it did not help.

### Solution

All in `Source/KineticMonteCarlo/CD_KMCSolverImplem.H` (with the header docs updated to match):

1. **SSA decision before the critical draw.** `advanceHybrid` now forms the leap candidate `dtLeap = min(remaining, dtNonCrit)`, which depends only on the state, and tests `A * dtLeap < SSA_lim` on it. The waiting time `dtCrit` is drawn only in the leap branch, after that decision, and is drawn afresh on every rejected retry, so no draw survives the step it belonged to. This is the ordering of the Cao–Gillespie–Petzold algorithm. The three-way branch (`useSSA` / `nonCriticalOnly` / critical) collapses into two.
2. **Gross-consumption bound.** `getNonCriticalTimeStep` accumulates, per reactant, the expected consumption `lambda_i = sum_{j: nu_ij < 0} |nu_ij| a_j` (new `m_lossScratch`, same pattern as `m_muScratch`/`m_sigmaScratch`) and adds the bound `tau <= f_i / lambda_i`. For a reactant that is only consumed it coincides with the existing mean bound, so it only bites when production cancels loss.
3. **Cao's species set.** The bounds are now taken over the reactants of *all* reactions (the paper's `I_rs`), with the sums still over the non-critical ones, and the floor `f_i = max(eps X_i, 1)` applies at `X_i = 0` instead of skipping the species. A species produced from zero is thus bounded to one expected net change per leap and the `SSA_lim` switch takes over. `getNonCriticalTimeStep` gained a leading `a_reactions` argument for this; there are no callers outside the solver.
4. **Rounding in `Random::getPoisson`** (`Source/Utilities/CD_RandomImplem.H`). Found by the convergence study below, and independent of the hybrid algorithm: for means of 250 and above the sampler draws a normal distribution and converted it to an integer with a plain cast, which truncates toward zero and biases every such draw by −1/2. Every leap propagator draws its firing counts through this function, so any reaction expected to fire more than 250 times per substep fired half a time too seldom. It now uses `llround`. Measured on 5M draws: the cast biases the mean by −0.50 at mean 300 and −0.59 at mean 10⁵; `llround` is within noise at the same cost (28 vs 26 ns per draw). An exact `std::poisson_distribution` would be ~105 ns per draw, so the approximation is kept.

Results on the reproducer (`Exec/Convergence/KineticMonteCarlo/C3`; it sits with C1 and C2 as a verification case with an exact reference, and has its own subsection and figure in the Sphinx verification section. 12 ranks × 20000 realizations, sampling error ≈ 0.2 %, exact mean 993.89):

| configuration | before (issue) | after |
|---|---|---|
| `ssa` | +0.10 % | +0.09 % |
| `hybrid_midpoint`, `max_dt = 1e-11` (default) | **−7.20 %** | −0.13 % |
| `hybrid_midpoint`, `max_dt = 1e-12` | +0.12 % | −0.05 % |
| `hybrid_midpoint`, `crit_num = 500` | **−66.16 %** | −0.38 % |
| … `SSA_lim = 0` | −0.17 % | +0.20 % |
| … `SSA_lim = 1000` | +0.27 % | −0.01 % |
| `hybrid_prc` | −4.3 % | −0.18 % |
| `hybrid_implicit_euler` | +7.2 % | +0.28 % |
| `hybrid_explicit_euler` | −7.1 % | −2.7 % |
| `prop_eps = 0.1`, `SSA_lim = 2` | −5.16 % | +0.09 % |

With 100000 realizations per rank the default configuration is at −0.004 %. `hybrid_explicit_euler` keeps a first-order truncation error (midpoint and PRC are second order), and `SSA_lim = 0` keeps −1.6 %: with the SSA disabled, the `f >= 1` floor still leaves the critical propensities frozen for at least one expected firing per leap, and the algorithm relies on the SSA switch there, as in the paper.

**Convergence study.** The C3 driver prints the mean populations after every step against the moment equations (which it integrates from the rates in `chemistry.json`), and takes the initial number of electrons as an input, so both limits can be checked: many realizations, and many initial particles (where tau-leaping rather than the SSA does the work). All numbers below are on the final code in this PR. With one initial electron and 12 × {200, 2000, 20000, 200000} realizations the deviation of `hybrid_midpoint` is −0.46, −0.46, +0.02, +0.002 % against a standard error of 1.86, 0.59, 0.19, 0.06 %: within one sigma throughout while the error falls as 1/√n. Against the initial population, with the pre-fix solver built from upstream `main` into a second binary and run through the same driver (12 × 2000 realizations):

| `N0` | pre-fix `hybrid_midpoint` | fixed `hybrid_midpoint` | pre-fix `hybrid_explicit_euler` | fixed `hybrid_explicit_euler` | `ssa` | stderr |
|---|---|---|---|---|---|---|
| 1 | −7.01 % | −0.14 % | −11.2 % | −2.2 % | +0.62 % | 0.6 % |
| 10 | −4.06 % | −0.11 % | −13.7 % | −4.2 % | +0.17 % | 0.19 % |
| 100 | −0.86 % | +0.07 % | −17.4 % | −5.9 % | −0.03 % | 0.06 % |
| 1 000 | −1.19 % | −0.06 % | −20.0 % | −6.1 % | +0.03 % | 0.02 % |
| 10 000 | −1.05 % | −0.04 % | −20.3 % | −6.1 % | +0.06 % (0.04 %) | 0.006 % |
| 1, `prop_eps = 1.E99` | −4.49 % | +0.81 % | | | | 0.6 % |
| 1 000, `prop_eps = 1.E99` | −1.17 % | −0.55 % | | | | 0.02 % |

At `N0 = 1000` the other propagators give `hybrid_prc` −0.10 % and `hybrid_implicit_euler` +6.7 %. The second-order `hybrid_midpoint` and `hybrid_prc` are thus within noise of both the SSA and the moment solution at every `N0`; the few-sigma remainder at `N0 ≥ 1000` (−0.04 to −0.06 %) is the size of the midpoint's own O(τ²) leap error, `(λτ)³/6` over ~345 substeps ≈ 0.05 %, and shrinks with `prop_eps`. The ∓6 % of the two Euler propagators at large `N0` is their first-order truncation error, `(1 ± x)` against `e^x` with `x = λτ ≈ 0.02` over ~345 substeps, which predicts ∓6.9 %; it shrinks linearly with `prop_eps` and is a property of the propagator, three times smaller than before. Wall-clock on this deliberately stiff chemistry goes from 0.5 s to 0.8 s, since the consumption bound turns one leap per step into about four.

Documentation: the hybrid-algorithm steps in `Docs/Sphinx/source/Solvers/KineticMonteCarlo.rst` are rewritten to the new ordering, a "Leap condition" subsection states the three bounds, and the `literalinclude` anchor into `CD_KMCSolver.H` is repointed. The `prop_eps` comment in `CD_ItoKMCJSON.options` now says it also controls reactant consumption. Note that `prop_eps` scales the bounds but never removes them: for `prop_eps >= 1` the expected consumption of a reactant may reach or exceed its population, so the bounds no longer keep the sampled firings below the populations and validity then rests on step rejection; the `f = 1` floor at zero population applies for any `prop_eps`. The issue's suggestion of a runtime warning on turnover was not implemented: after the fix the turnover per substep is bounded by `prop_eps` by construction, so the warning could only fire when `prop_eps >= 1`, where it would fire per cell per step.

### Side-effects

- **Benchmark files will change.** Fix 1 alters the random-number consumption of every `hybrid_*` run, fix 3's zero-population floor takes effect even with `prop_eps = 1.E99` (one extra short substep whenever a reactant species is produced from zero), and fix 4 changes every Poisson draw with a mean of 250 or more, in every tau-leaping algorithm, hybrid or not. No `tests.py --compare` pass was run; the existing `KineticMonteCarlo/Avalanche` and `Schlogl` tests build and run.
- `KMCSolver::getNonCriticalTimeStep(state, nonCritical, propensities)` becomes `(state, reactions, nonCritical, propensities)`. Only internal callers exist.
- One new member on `KMCSolver` (`m_lossScratch`), a mutable scratch vector alongside the existing three. No new types.
- Cost: `getNonCriticalTimeStep` now reduces over the reactants of all reactions rather than the non-critical subset, and `advanceHybrid` evaluates the total propensity once per substep instead of once per retry. In the small-population regime a species produced from zero now forces short substeps (and hence the SSA) until it is populated, which is the CGP cost model; for `prop_eps = 1.E99` this is a single extra substep.

### Alternative solutions 

- For fix 1, the issue's other suggestion was to apply the pending critical firing in the SSA branch as well. Deciding before the draw is what the paper does and removes the coupling entirely, so it was preferred.
- For fix 2, the issue proposes the per-reaction bound `a_j tau <= eps L_j`. The per-species consumption bound implemented here sums the consumers of each reactant and therefore implies the per-reaction one; it also slots into the existing per-species accumulation loop with no extra reaction-interface calls.
- Making the fast reaction critical based on rate (the issue's "related" remark on `partitionReactions`) was not needed once the consumption bound resolves it as a non-critical reaction.
- Raising the default `SSA_lim` to the paper's 10 would have masked fix 3 (it does on the reproducer) but not addressed it.

### PR-review checklist

Mandatory:

Optional:

https://claude.ai/code/session_01UEDMxsTzKgFRSukSAsG7jB


rmrsk:
@claude review

rmrsk:
@claude review

PR #730, opened 2026-09-11
Generate cut-cell moments from an explicit polyhedron

rmrsk:
# Summary

> Step 1 of the roadmap in #733. Merge order is **#730 → #731**; #732 is rebased on top later.
>
> - **#730 (this one)** — the polyhedral moment machinery: a cut cell's moments become exact integrals of a closed polyhedron rather than quadratures of the implicit function, and a cell is classified by one rule from its own corner values.
> - **#731** — the AMR-aware grid builder in `ComputationalGeometry`, the polyhedral surface reconstructed over those grids with coarse–fine seams, and its watertightness check.

Adds a third geometry generator, `Driver.geometry_generation = polyhedral`. The default is unchanged, so nothing moves unless the switch is thrown.

### Background

Chombo stores moments and never forms a surface. Anything that wants to subdivide a cut cell therefore has to *reconstruct* geometry from an over-determined moment set, and every reconstruction we tried carried its own tail of failures — a plane fitted to the stored moments leaves 3.7–12% of children covered but with open faces; a quadric is only safe for one curvature sign; a least-squares system over the child moments has 107 free dimensions against a realisability constraint that is not convex.

Defining the moments as the moments of a stored body removes the problem rather than solving it: subdivision becomes exact clipping, and every coarsening relation becomes an identity about a partition instead of a condition to satisfy. That is what #731 and #732 build on.

### Solution

`PolyhedralGeometryShop : public ScanShop` overrides `fillGraph`. It classifies every cell from its own corner values, reuses `GeometryShop::fixRegularCellsNextToCovered` for the regular cells bordering a covered one, bisects each cut edge once through a cache so the four cells sharing it agree bit for bit, and builds each node from the polyhedron. It never calls `computeVoFInternals`, so `Moments.cpp` and the least-squares machinery are skipped entirely.

**Classification is one predicate.** A corner value is fluid when it is negative and solid otherwise, an exact zero of either sign included; a cell is cut as soon as two corners disagree, regular if all are fluid, covered if all are solid. `GeometryShop` reads an exact zero by its sign bit — which a negated distance function turns over, so the same node-plane face comes back as plus zero, minus zero and a few parts in 1e19 at neighbouring nodes — and then rescues the regular cell that leaves beside a covered one by looking at its neighbour. Here the zero falls on the solid side by rule, so a fluid cell whose face lies in the interface is a cut cell with κ = 1 and that face as its interface, and the cell across it is covered; nothing has to be repaired afterwards. To make the rule mean something, a value within 1e-12 of a cell width of zero is read as exactly zero (`snappedValue`), in the node values and in the bisection alike.

**Roots go through zeros.** A corner at exactly zero is no longer taken to be the crossing: the function may be zero along a stretch of the edge, and `edgeRoot` bisects through the zeros to where it leaves them. A root within 1e-4 of an edge from an endpoint that is exactly zero is put on that endpoint, which is what keeps a surface tangent to a node from producing a sliver. The mirror image of the covered threshold is a dust rule: a body with neither solid volume nor boundary area is a regular cell.

The cut cell lives in `PolyhedralEB`: `CutCellSurface` holds the corner values and edge crossings, `CutCellBody` assembles the face polygons and the interface patch from them and takes the moments. The helpers in `CD_PolyhedralEBUtils.H` are the one place the corner, edge and face numbering is defined; the shop decodes edges through them rather than carrying its own copy of the convention. The volume-fraction threshold drops cells exactly as `GeometryShop` does — over the valid region only, since the same `a_validRegion` filter feeds `ivsirreg` there — and the dropped-cell fix-up finds a neighbour's node through a cell-indexed `BaseFab<int>` rather than by scanning the node list, which `GeometryShop` does and which is quadratic in the number of cut cells in a patch.

Every invariant the code relies on is a `CH_assert`: edge, corner, face and polygon bounds; the one-polygon body in two dimensions; the single covered face of a regular cell; the pairing `facePairs` promises `faceWalk`; the ghost region the node filling reads across. A `DEBUG` build checks them all; an optimised build compiles them out.

Compared against the existing generator, same runs, cell by cell. **Topology differences are zero everywhere.** These numbers predate the classification rule above; the curved and oblique rows are unaffected by it (no corner is exactly zero there), but the grid-aligned rows now differ from `GeometryShop` *by design* — a face lying in a node plane is a cut cell with κ = 1 rather than a regular cell next to a covered one — and are to be re-measured before merge.

| geometry | cells | topology | Δ volFrac | Δ areaFrac |
| --- | ---: | ---: | ---: | ---: |
| oblique plane 3-D | 8,898 | 0 | 3.8e-15 | 8.6e-15 |
| sphere 3-D | 1,220 | 0 | 4.9e-4 | 8.4e-2 |
| torus 3-D | 4,208 | 0 | 2.3e-3 | 7.8e-2 |
| rotated cube 3-D | 6,307 | 0 | 5.8e-2 | 4.9e-1 |
| rotated simplex 3-D | 9,595 | 0 | 1.5e-1 | 5.2e-1 |
| axis-aligned simplex 3-D | 9,852 | 0 | 1.2e-1 | 3.4e-1 |
| axis-aligned cube 3-D | 4,376 | 0 | **0.00e+00** | **0.00e+00** |
| oblique line 2-D | 105 | 0 | 3.6e-15 | 5.8e-15 |
| grid-aligned line 2-D | 64 | 0 | **0.00e+00** | **0.00e+00** |
| axis-aligned box 2-D | 108 | 0 | **0.00e+00** | **0.00e+00** |
| rotated box 2-D | 104 | 0 | 2.2e-15 | 2.2e-15 |
| circle 2-D | 64 | 0 | 2.3e-12 | 4.5e-12 |
| grid-aligned circle 2-D | 60 | 0 | 6.5e-13 | 1.2e-12 |

Planar geometry agrees to rounding and grid-aligned planar geometry agrees exactly. Curved geometry differs by the third-to-second-order aperture trade, which is the one deliberate accuracy cost. All of 2-D agrees to rounding, because there the interface inside a cut cell genuinely *is* the chord between its crossings.

Sphere at 64³: **0.043 s** against **0.068 s**. Root finding costs the same either way; the difference is the moment quadrature this never runs.

`Electrostatics/MechShaft` with plain multigrid needs 10/11 iterations in two dimensions and 15/15 in three against the existing generator; `ProfiledSurface` with BiCGStab needs 4/4 and 5/5.

**Regression suite**, both dimensions, benchmarks generated from the pre-change tree:

| | 2-D | 3-D |
| --- | --- | --- |
| comparisons | 87 | 87 |
| byte-identical, this PR's code with the submodule pinned back | **87** | **87** |
| byte-identical, with the Chombo fix in | 86 | 86 |

The single difference in each dimension is `Electrostatics/MechShaft`, and it comes from the Chombo submodule bump rather than from anything here. This comparison predates the `RoundedCylinderSDF` fix listed under side-effects; `Geometry/SpherePlane` changes with that fix and its benchmark needs regenerating from this tree.

The pin moves to Chombo-3.3#30, a separate merged fix: `fixRegularCellsNextToCovered` was losing the boundary centroid on every regular cell bordering a covered one that an earlier covered cell had already claimed by edge or corner. On an axis-aligned cube, 4,050 of 4,056 cells wrong, with exactly six correct — one per face of the cube, being the cells whose covered partner comes first in `BoxIterator` order.

**New types**, with use counts across `Source/`, `Physics/` and `Geometries/`:

| type | kind | uses |
| --- | --- | ---: |
| `PolyhedralEB::CutCellSurface` | POD: corner values and edge crossings of one cell | 70 |
| `PolyhedralEB::CutCellBody` | the polyhedron and its moments | 40 |
| `PolyhedralEB::CutCellBody::Polygon` | one face or interface polygon of the body, in the cell frame | 15 |
| `PolyhedralEB::CutCellBody::Kind` | covered / regular / cut classification | 18 |
| `PolyhedralGeometryShop` | the generator | 30 |
| `ComputationalGeometry::Generator` | which of the three generators the input file selected | 12 |

Each is a genuinely new domain concept rather than a renaming of something that existed; there is no map or hand-rolled per-cell store, the node index being a `BaseFab<int>` over the valid region. The utilities also name the vertex weld tolerance (`s_weldTolerance`, `sameVertex`) the polygon machinery matches vertices by.

### Side-effects

- **The default generator is unchanged**, so existing benchmarks stay valid. `polyhedral` must be asked for.
- **`polyhedral` runs differ from `chombo-discharge` runs by design** on curved geometry. They are not interchangeable for benchmark comparison; that is the point of the switch.
- **The submodule pin moves.** Chombo-3.3#30 changes stored output for any geometry with facets lying exactly on a grid plane — an axis-aligned electrode face is the common case. This is a material change, not a last-bit one: on `MechShaft` the values still differ at a 1e-1 relative tolerance. The old values were wrong. Benchmarks for such setups need regenerating from this tree.
- **A face lying in a node plane is a cut cell.** Under `polyhedral`, an axis-aligned electrode face gives cut cells with κ = 1 on the fluid side and covered cells on the solid side, where `GeometryShop` gives regular cells beside covered ones. Solvers see the same fluxes; the index space, the irregular-cell sets and plot files differ for such geometries.
- **Two tolerances are introduced**, both fixed in the source rather than input options: values within 1e-12 of a cell width are exact zeros, and a root within 1e-4 of an edge from a zero endpoint sits on it. Neither scales with the geometry; both scale with the cell.
- **A cell whose body will not close aborts the run.** It has not fired on any geometry tested. The check is `ComputationalGeometry::s_strictGeometry = true`, deliberately *not* an input-file option: turning it off is an edit to the source, so it cannot happen by accident. The Sphinx page says the same.
- **`GECReferenceCell` and `SpherePlane` change shape.** Since the EBGeometry updates on `main`, `RoundedCylinderSDF`'s first argument is the outer radius, and these two geometries were still passing `0.5 * r + curvature`. They now pass `r`, so the vessel and inner electrode of `GECReferenceCell` and the disk of `SpherePlane` are the size their input files say. This is independent of the generator switch. `Geometry/SpherePlane` is in the regression suite and its benchmark changes; `GECReferenceCell` is an example, not a test.

### Alternative solutions

Modifying `GeometryShop` directly was considered and rejected: carrying a parallel generator allows A/B comparison against Chombo on identical grids, and leaves a known-good path to fall back to while debugging. It turned out to cost nothing, because `GeometryShop.H` is public exactly through the functions a subclass needs — `fixRegularCellsNextToCovered` and `getFullNodeWithCoveredFace` carry the subtle covered-cell logic and are reachable, while the private ones are precisely the parts the prototype already reimplements.

Harvesting Chombo's edge roots by calling `computeVoFInternals` and discarding its moments was considered and rejected: it would roughly double generation time to buy bit-identical roots, which only matter if Chombo's roots and ours are mixed — they are not.

The prototype exporter and the standalone C++ spike every measurement came from were carried on the branch for a while and then left out of this change; they live outside the repository.

### PR-review checklist

Mandatory:

Optional:


rmrsk:
@claude review

PR #731, opened 2026-09-11
Generate single-valued cut cells over AMR grids

rmrsk:
# Summary

> Step 2 of the roadmap in #733, on top of #730 (merged). #732 is rebased onto this afterwards.
>
> No Chombo change. The submodule pointer is `main`'s.

`ComputationalGeometry` builds the AMR grids the cut-cell geometry wants — one box hierarchy for both phases, refined where the interface turns and nowhere else, properly nested, classified on every level — before any index space exists. `PolyhedralGeometryShop` reconstructs the polyhedral surface over those grids, closes the coarse–fine seams, checks the result is watertight and can write it as STL. Under `Driver.geometry_generation = polyhedral` the simulation's initial AMR grids are these grids.

The index space is *not* yet generated over them: the moments are still computed by the polyhedral shop over the full grids it inherits from `ScanShop`, exactly as in #730. That is the next step (#733, step 6).

### Background

Geometry generation today resolves every cut cell at the finest spacing, whether the surface there is curved or flat, and the simulation's refinement of the geometry (`Driver.refine_angles`) is decided afterwards from the EB normals of the finished index space. The point of the roadmap is to spend resolution where the surface needs it, which requires deciding the grids first, from the implicit functions alone, and requires them to be (a) refined only where the surface turns, (b) properly nested, (c) classified regular, covered or irregular on every level, and (d) the same boxes for both phases. `ScanShop` does none of this. An earlier form of this PR tried to bolt partial coverage onto `ScanShop`'s levels after the fact; the design record of what that ran into, and why the grids are now built first, is kept with the roadmap (#733) rather than in the tree.

### Solution

**`ComputationalGeometry::makeGrids(startDomain, stopDomain, probLo, startDx, refineAngle, maxGhostEB)`**, called from `Driver::buildGeometryGrids` before `buildGeometries` on all three setup paths, under `polyhedral` only. The stop domain and its spacing are read from the AMR level arrays at `max_amr_depth`, so the level the geometry is resolved on can later become `max_eb_depth` without touching this.

1. **Start level.** `domainSplit` of the start domain (the scan domain); each box classified in both phases by `ScanShop`'s cell-centre test over the box grown by the EB ghost width.
2. **Upward pass to the stop domain.** A box regular or covered in both phases refines whole. A box irregular in some phase is tested for curvature in that phase and split into classified pieces, or left as a leaf with nothing above it. The pass always runs to the stop domain (the finest AMR domain).
3. **Cut tiles.** One `TiledMeshRefine` over the boxes irregular in either phase gives a properly nested, tile-aligned set common to both phases (`m_cutTiles`, `getCutTiles`), with a one-tile nesting buffer. The tile is `ComputationalGeometry.min_block_size` and the super-tile `ComputationalGeometry.max_block_size` (both 8 unless given) — deliberately the geometry's own and not `AmrMesh`'s, since the refined footprint beside a curved feature is the parent box plus one nesting tile per level.
4. **Tile classification.** Each tile lies in one box or above a leaf (start-level lattice arithmetic plus the pass's child links); per phase it inherits a regular or covered host's tag and is otherwise classified by that phase's implicit function at its own level.
5. **Decimation.** Each hit regular or covered box is cut down to what the tiles left of it (`TreeIntVectSet` in tile units; octree-graded, not tight, replaceable in one place).
6. **Coarser levels and push-down.** The levels below the start domain, whole; then, finest to coarsest, a box containing a box irregular in a phase is irregular in that phase.

Classification and the curvature test are shared across ranks by index and assembled with one all-reduce per list; every rank ends with identical lists. `makeGrids` reports per-level box and tile counts, how many boxes split for which reason, and a `Timer` breakdown of its six steps to `pout`.

**The curvature test** compares the interface normals of neighbouring cut cells (the 3^D block, one cell beyond the box so a pair across the boundary is seen from both sides) against `Driver.refine_angles`; a pair inside the box splits it, a pair reaching into the ring splits it only if it has no pair of its own, and facing normals (dot < 0, a solid or a gap thinner than two cells) split it regardless of the angle. The normals are those of the interface polygons the shop reconstructs — the same `fillNodeValues`/`buildSurface` the graph uses, now statics over an explicit implicit function, with node values once per node of the box and every crossed edge bisected once into the shared edge cache — and depend on the bisected edge roots alone. This matters: the implicit function is not a distance function inside a body built by CSG, and its gradient there carries the kinks and medial shells of the construction — measured on `ProfiledSurface`, `f` under the wheel's tread rises to the torus tube's axis, falls and rises again, so a gradient-based band test refined a 31 mm cylinder on every level, patchily along its circumference. The polygon normals do not see any of that.

**The surface on those grids.** `PolyhedralGeometryShop::setGrids` / `verifySurface` reconstruct every cut cell of every irregular box at its own level from one node-value fab and one edge cache per box, from the finest level that carries each cell, and where a cell borders a region the next finer level carries, restrict that face to the finer level's chords and close the interface (`CutCellBody::restrictFace`, `closeInterface`, `mergeCoplanar`; three dimensions). `sanityCheck` (`PolyhedralGeometryShop.sanity_check`, default true) welds the vertices at 1e-8 of the finest spacing and requires every interior triangle edge to be used exactly twice, aborting otherwise; `PolyhedralGeometryShop.write_stl` (default false) writes `surface_mesh_<phase>.stl`, one per level, and the boxes and split boxes per level as text. `verifySurface` reports a `Timer` breakdown per level, STL and check.

**Simulation grids.** `AmrMesh::regridAmr(const Vector<Vector<Box>>&, int)` takes the boxes of every level instead of generating them from tags; the load balancing and `DisjointBoxLayout` definition it shares with the tag path are factored into `defineGrids`. `Driver::getGeometryGrids` maps the geometry's cut tiles onto the AMR levels and the initial regrid uses them under `polyhedral`; later regrids and restart still go through tags.

**Verified** on `Exec/Examples/Electrostatics/ProfiledSurface` in 3-D, 16 ranks, geometry only, `polyhedral`, depth 3, 15°, tile 4: both phases watertight (1 612 260 gas and 1 384 604 solid triangles, 0 open, 0 over-used), AMR box counts per level equal to the cut-tile counts. Timings from the same run: `makeGrids` 1.0 s (0.86 s of it the upward pass), the polyhedra over all levels 0.51 s gas and 0.45 s solid, the sanity check 1.3 s and 1.0 s on rank 0. Geometry time is 9.0 s with the check and 6.4 s without, of which ~4.4 s is the index space over `ScanShop`'s full grids; the `chombo-discharge` generator takes 11.7 s on the same case. The committed `example3d.inputs` (`chombo-discharge` generator, full solve) runs to completion. Compiles and links in 2-D and 3-D. The regression suite has not been run on this branch; the `chombo` and `chombo-discharge` paths are unchanged apart from where the implicit functions are built.

**New types**: `ComputationalGeometry::SplitReason` (enum, why a box split; 4 uses in the report). No map, no hand-rolled per-cell store; the per-cell fields of the curvature test are `BaseFab`s over the grown box.

### Side-effects

- **Nothing changes for the `chombo` and `chombo-discharge` generators.** The builder and the surface pass run only under `polyhedral`. The one shared change is that the two phases' implicit functions are built once, in `buildImplicitFunctions`, ahead of the generators — the same functions as before.
- **Under `polyhedral` the index space is still `ScanShop`'s full grids** with the polyhedral moments on them. The builder's grids drive the initial AMR grids and the surface check.
- **Under `polyhedral` the initial AMR grids are the geometry's tiles**, sized by `ComputationalGeometry.*_block_size` rather than `AmrMesh.*_block_size`. Fine for `geometry_only`; a solve on those grids needs the two block sizes reconciled, which is the regrid step of the roadmap.
- **`sanity_check` is on by default** and costs a reconstruction of every cut cell on every level plus a gather of the whole surface on rank 0 (about 150 bytes per triangle there). It also aborts on a geometry the seam cannot represent — a coarse edge the finer level crosses twice, a real sub-cell grazing crossing; `ProfiledSurface` with `wheel_extra_thickness = 1E-3` at 45° is such a case. Set `PolyhedralGeometryShop.sanity_check = false` to run anyway. Whether the default should stay on once the graph makes the check local is an open question.
- **New inputs**: `ComputationalGeometry.min_block_size`, `ComputationalGeometry.max_block_size`, `PolyhedralGeometryShop.sanity_check`, `PolyhedralGeometryShop.write_stl`. None is documented in Sphinx yet.
- **Preconditions the builder aborts on**: `min_block_size ≥ 2 × EB ghost cells`, `max_block_size` a multiple of `min_block_size`, the start domain decomposing into whole tiles, the stop domain a factor-two refinement of the start domain.
- **Three Sphinx `literalinclude` ranges moved** with the header additions (`ComputationalGeometry.rst` twice, `Particles.rst` once); every other range into the touched files was checked and is unchanged.

### Alternative solutions 

- **Partial coverage grafted onto `ScanShop`'s levels.** The first form of this PR; could not keep the grids properly nested, left holes that read back as regular fluid, and needed the index space extended in place across regrids. Recorded in the design notes kept with #733.
- **One hierarchy per phase, unioned afterwards.** The tiles come out common either way, but the per-phase holes differ and a tile from the other phase lands in a hole the phase has to reclassify. Building in tandem from both implicit functions removes the union and half the storage.
- **Gradient normals in a band around the zero set** for the curvature test. That was the first version; see above for why it was replaced by the reconstructed polygon normals.
- **Tagging cells rather than splitting boxes.** Would make the refined footprint one tile around the sharp cells instead of the parent box; it changes the builder from box-driven to tag-driven and is left for the hardening step, with the small geometry tile as the interim answer.
- **A per-level BVH now.** The lookups the build needs follow from the lattice and the pass's child links; a BVH is what lets the geometry answer for a domain finer than anything stored, and is step 4 of the roadmap.

### PR-review checklist

Mandatory:

Optional:


rmrsk:
@claude review

rmrsk:
@claude review

rmrsk:
@claude review

rmrsk:
@claude review

PR #732, opened 2026-09-11
Support polyhedral refinement

rmrsk:
# Summary

> **Draft — do not merge.** Third and last of a stack. Merge order is **#730 → #731 → #732**, and this is #732. It is branched from #731 and so currently shows that work and #730's as well; it **will be rebased** once the two below it land, at which point it reduces to the polyhedral refinement itself.

Refining a single-valued cut cell by cutting the body its moments are taken from, so that the pieces partition the parent exactly rather than being separate reconstructions of the same surface.

### Background

#730 makes a cut cell's moments the exact integrals of a closed polyhedron. The reason for doing that was never the moments in themselves — it was that a body can be *cut*, and a set of moments cannot.

Refining a cut cell whose moments are all one has means reconstructing geometry from an over-determined set, and every reconstruction tried during design carried its own tail of failures: a plane fitted to the stored moments left 3.7–12% of children covered but with open faces, a quadric is safe for only one curvature sign, and a least-squares system over the child moments leaves 107 free dimensions against a realisability constraint that is not convex.

With a body in hand, refinement stops being a reconstruction and becomes a partition, so every coarsening relation is an identity rather than a condition to satisfy.

### Solution

`CutCellBody::refine` cuts the body with the cell's own mid-planes and caps each opening, producing the cells that partition it. Checked against the parent over **38,124 cut cells and all 304,992 children** on six geometries in three dimensions and three in two:

| relation | 3-D | 2-D |
| --- | ---: | ---: |
| volume fraction coarsens back | 4.4e-16 | 1.1e-16 |
| apertures | 2.2e-16 | 0 |
| boundary area vector | 4.4e-16 | 2.2e-16 |
| true boundary area | 6.7e-16 | 2.2e-16 |
| children close | 5.0e-16 | 5.6e-17 |

No cell refused, no volume fraction or aperture out of range, no centroid outside its own cell.

`Driver.geometry_refinement` says how many times the generated cells refine the cells whose surface is reconstructed: at 1, the default, each cell's surface is reconstructed on the cell itself; at 2 it is reconstructed on the parent and cut in half along each direction; higher powers of two cut repeatedly.

Classification is inherited rather than redone. A refined cell asks the piece of the parent body it received — no polygons means covered, polygons lying only in cell faces mean the interface never enters it, anything else means cut — so there is one classification, made where the surface is reconstructed and where single-valuedness is mandated. A coarse cell the interface never enters answers for all of its children without a body being built.

On a sphere the cut-cell count goes 1526 → 1518 → 1476 at refinements 1, 2 and 4, which is the sphere being represented on a coarser surface each time. At every refinement the generated levels satisfy `Σ(α_hi − α_lo) = a_B n` to 3.5e-16, with nothing out of range, no isolated or ghost cells, and no multi-valued cells.

### Also owned here: completing a partly carried index space

#731 lets the index space be carried only where the geometry needs resolving. That leaves it short of the contract the rest of Chombo relies on for refinement, and closing that is this PR's job rather than #731's, because the operation it needs is the one this PR is about.

`GraphNode::refine` is documented as *"the result is only defined if this EBGraph was defined by coarsening"*. In an ordinary index space that always holds — every level below the finest is the coarsening of the one above it, over the whole domain — so callers may refine any coarse cell they can see. A level carried in part is defined by coarsening only where the finer level reaches.

The capability is promised level-wide, not merely across a refinement boundary: `PhaseRealm::defineEBLevelGrid` calls `setMaxRefinementRatio` on every level below the finest, and that is redeemed in `EBISLayoutImplem::setMaxRefinementRatio`, which refines the whole coarse layout and calls `fillEBISLayout` on the finer level. Today that always succeeds because the finer level spans the domain. Under partial coverage it asks for finer-level graph and data where none was carried, and `EBCoarseFineParticleMesh::defineStencils` is the first client to notice.

`CutCellBody::refine` is exactly what fills it: where the index space carried no finer level, the refined layout is built by cutting the coarse bodies into the children that partition them. The coarsening relations are already verified as identities to ~1e-16 over 304,992 children, which is the property the filled layout has to have. The work is in Chombo, in `fillEBISLayout` and what it reaches.

### Side-effects

- **Default behaviour is unchanged.** `Driver.geometry_refinement` defaults to 1 and is read with `query`, so input files written before it keep working. It is reachable only through `Driver.geometry_generation = polyhedral`, itself not the default.
- **Moments at refinement > 1 differ from generating at the fine resolution directly**, and must: a coarser reconstruction of a curved surface is a different surface. On a sphere the volume fractions differ by up to 1.1e-1, and two cells gain a face the fine reconstruction does not cut. Both cells sharing that face agree about it from either side in both cases, which is the property that has to hold.
- **The coarse body is rebuilt once per child.** At refinement 2 in three dimensions the coarse surface is reconstructed nine times per coarse cell. Generation is a one-off and this is opt-in, but the cost grows with refinement and is worth revisiting.

### Alternative solutions

Cutting level by level and flattening the pieces into one array was tried and abandoned: the flattening order stops matching `BoxIterator`'s beyond one refinement, and the array would have to hold one entry per cell of the whole refinement. Descending per cell re-cuts the parent once per child but is unambiguous, which is the cost noted above.

Classifying a refined cell from its own corner values was the first implementation and is wrong: it leaves two resolutions answering one question, and a cell the fine corners call cut can receive an empty piece of its parent's body.

### PR-review checklist

Mandatory:

Optional:


rmrsk:
Closing because of a new roadmap that makes this PR obsolete.

